PR #2, opened 2015-01-19
DM-245: Implement HSC camera in new camera framework

PaulPrice:
Pull request for DM-245, from u/lauren/DM-245 .  This is the branch I'm *supposed* to review, not tickets/DM-245.


PaulPrice:
We need to check that we can process and get decent results with HSC, SC and SC-MIT.

PR #3, opened 2015-07-27
Review for DM-1794

jdswinbank:


jdswinbank:
Unfortunately, "we could not display the entire diff because it was too big", which makes it impossible to provide meaningful comments here. I'll use JIRA instead.

timj:
Can we mark FITS files as binary files using gitattributes so Github doesn't try to show us the diff. Plain text merging of changes to FITS files seems risky in general.

PR #4, opened 2015-08-13
obs_subaru DM-1954: HeavyFootprints in forced coadd photometry

TallJimbo:


config/forcedPhotCcd.py
```diff
@@ -2,8 +2,10 @@
 
 root.measurement.load(os.path.join(os.environ['OBS_SUBARU_DIR'], 'config', 'apertures.py'))
 root.measurement.load(os.path.join(os.environ['OBS_SUBARU_DIR'], 'config', 'kron.py'))
-root.measurement.load(os.path.join(os.environ['OBS_SUBARU_DIR'], 'config', 'cmodel.py'))
+# Turn off cmodel until latest fixes (large blends, footprint merging, etc.) are in
+# root.measurement.load(os.path.join(os.environ['OBS_SUBARU_DIR'], 'config', 'cmodel.py'))
 
 # Disable Gaussian mags (which aren't really forced)
 root.measurement.algorithms.names -= ["base_GaussianFlux"]
+root.measurement.algorithms.names |= ["base_SdssShape"]
```
TallJimbo: I don't think we should be running `SdssShape` in forced mode, either here or in {{forcedPhotCoadd.py}}, below; it's not actually a forced measurement, and it doesn't look like we're running it there on the HSC side.  Did you find that you had to run it for some reason, or is my assertion that we're not running it on HSC incorrect?

And actually, It think we can remove the line above (and again in `forcedPhotCoadd.py`), since we've fixed `GaussianFlux` on the LSST side such that it actually *does* do forced photometry (it's not PSF-corrected, but neither are the aperture magnitudes).

TallJimbo:
I don't think we need the `forcedCcd.py` and `forcedCoadd.py` config files anymore (if we do, it means we're inconsistent somewhere in what we call these tasks, and that should be fixed).  I see that they're present on the HSC side as well, but I think they're similarly useless there.

config/forcedPhotCcd.py
laurenam: Oh, I added that to have more measurements to compare with in the forced vs. meas catalogs.  I'll remove it and the one above in both files.

PR #6, opened 2015-08-30
Review of DM-3493 (Fix crosstalk following ds9 interface changes)

jdswinbank:


python/lsst/obs/subaru/crosstalk.py
```diff
@@ -511,16 +513,16 @@ def fixCcd(butler, visit, ccd, coeffs, display=True):
     """Apply cross-talk correction to a CCD, given the cross-talk coefficients"""
     mi = readImage(butler, visit=visit, ccd=ccd)
     if display:
-        ds9.mtv(mi.getImage(), frame=0, title="CCD %d" % ccd)
+        frame1 = afwDisplay.getDisplay(backend="ds9", frame=1)
```
jdswinbank: Is it necessary to hard-code the `ds9` backend here? It should be the default unless the user has overridden it -- and if they have, we presumably want to respect their choices unless this code actually requires `ds9`.

jdswinbank: It looks as though we previously put the pre-subtraction image in frame `0` and the post-substraction in frame `1`. You seem to be putting both in `1`. Is that deliberate?

python/lsst/obs/subaru/crosstalk.py
```diff
@@ -226,50 +227,51 @@ def subtractXTalk(mi, coeffs, minPixelToMask=45000, crosstalkStr="CROSSTALK"):
...
+        # Clear the crosstalkStr bit in the original bright pixels, where tempStr is set
+        #
+        msk = mi.getMask()
+        temp = msk.getPlaneBitMask(tempStr)
+        xtalk_temp = crosstalk | temp
+        np_msk = msk.getArray()
+        np_msk[np.where(np.bitwise_and(np_msk, xtalk_temp) == xtalk_temp)] &= ~crosstalk
+
+    finally:
         msk.removeAndClearMaskPlane(tempStr, True) # added in afw #1853
```
jdswinbank: Given this has been added to `afw` for several years now, I agree that it's appropriate to drop the check for `AttributeError`. Should that be a separate commit, though? It's not really relevant to the conversion to the new `afwDisplay` interface.

PR #9, opened 2015-10-30
Review for DM-3942 (Preliminaries for the LSST vs. HSC pipeline comparison through single-frame processing)

jdswinbank:


python/lsst/obs/hsc/hscMapper.py
```diff
@@ -84,23 +91,28 @@ def __init__(self, **kwargs):
...
+            "NB0921",
+            "IB0945",
+            "NB0973",
             "SH",
             "PH",
             "NONE",
             "UNRECOGNISED"]:
             # Get the canonical name -- see #2113
             self.filters[f] = afwImage.Filter(afwImage.Filter(f).getId()).getName()
+        self.defaultFilterName = "UNRECOGNISED"
```
jdswinbank: Looks like this requires https://github.com/HyperSuprime-Cam/daf_butlerUtils/commit/daee24edba01b01a0412df7f9b4cf70be5b10860 which hasn't been merged to LSST.

jdswinbank: Discussed this with @laurenam; we agreed that it's ok to merge this, on the understanding that we still have to merge the rest of HSC-987 on a different (LSST) ticket and that it'll be easy to backtrack and take care of this if it causes problems in DM-2984.

python/lsst/obs/hsc/hscMapper.py
```diff
@@ -127,7 +139,12 @@ def map(self, datasetType, dataId, write=False):
         """
         copyId = dataId.copy()
         copyId.pop("flags", None)
-        return super(HscMapper, self).map(datasetType, copyId, write=write)
+        location = super(HscMapper, self).map(datasetType, copyId, write=write)
+
+        if 'flags' in dataId:
+            location.getAdditionalData().set('flags', dataId['flags'])
+
+        return location
```
jdswinbank: This is described as a fix for a bug in the Butler. Does the bug still exist in LSST? If so, do we have a ticket for it?

jdswinbank: Re-reading: I think the commit message is confusing here and that this probably isn't indicative of a bug in the `daf.persistence` Butler so much as a fix to the way `HscMapper` is interfacing with it. As such, the code is fine, but we should tweak the commit message.

laurenam: I agree.  History preservation question: should I change the commit message, or just add a note to it?

jdswinbank: I'd change it, since it's in the summary line: that's what people will see first when they read the logs, and we don't want them to be confused.

laurenam: Will do.

PR #11, opened 2015-12-22
DM-3518: Prune stale obs_subaru dependencies

jdswinbank:


ups/obs_subaru.table
```diff
@@ -1,19 +1,13 @@
...
 
-setupOptional(psycopg2)
-setupOptional(meas_extensions_multiShapelet)
-setupOptional(meas_multifit)
+setupOptional(meas_modelfit)
 setupOptional(meas_extensions_psfex)
 
-setupOptional(suprime_data)
-
-setupOptional(testdata_subaru)
```
timj: The tests still rely on `testdata_subaru`.

jdswinbank: It's a fair point. My initial reaction was to simply skip the tests on the assumption that they'll be covered [elsewhere](https://github.com/lsst/ci_hsc) pretty soon. On reflection, though, it's probably worth spending a few minutes putting together a minimal test data package and doing some bare-bones testing à la `testdata_decam`. Watch this space.

bin.src/registryInfo.py
```diff
@@ -178,11 +139,7 @@ def queryCalibRegistry(what, filterName=None, summary=False):
         print "%-10s--%-10s  %-10s  %-7s %-24s %4s" % (
             "validStart", "validEnd", "calibDate", "filter", "calibVersion", "nCCD")
 
-    if isSqlite:
         ret = cursor.execute(query, vals)
```
laurenam: Need to go back one indent level?

tests/hscRepository.py
```diff
@@ -0,0 +1,153 @@
...
+    @param[in] mapperName   Name of mapper class for use with repository.
+    @param[in] inputPath    Location of data files which will be ingested.
+    """
+    repoPath = mkdtemp()
+    with open(os.path.join(repoPath, "_mapper"), "w") as _mapper:
+        _mapper.write(mapperName)
+    ingest_cmd = os.path.join(getPackageDir("obs_subaru"), "bin.src", "hscIngestImages.py")
+    check_call([ingest_cmd, repoPath] + glob(os.path.join(inputPath, "*.fits.gz")))
+    return repoPath
+
```
laurenam: Add "action" to function name (e.g. createDataRepository)?

tests/hscRepository.py
```diff
@@ -0,0 +1,153 @@
...
+    def tearDown(self):
+        rmtree(self.repoPath)
+
+    def testIngest(self):
+        self.assertTrue(os.path.exists(os.path.join(self.repoPath, "registry.sqlite3")))
+
+    def testMapperName(self):
+        butler = dafPersist.Butler(root=self.repoPath)
+        self.assertEqual(butler.mapper.packageName, "obs_subaru")
+
```
laurenam: The setUp and tearDown actions here are repeated below.  Is a seperate class necessary/useful?

tests/hscRepository.py
```diff
@@ -0,0 +1,153 @@
...
+        ccd = raw.getDetector()
+        self.assertEqual(raw.getDimensions(), afwGeom.Extent2I(*self.rawSize))
+        self.assertEqual(raw.getFilter().getFilterProperty().getName(), self.filter)
+        self.assertEqual(ccd.getId(), self.ccdNum)
+        self.assertEqual(ccd.getBBox().getDimensions(), afwGeom.Extent2I(*self.ccdSize))
+
+    def testRawMetadata(self):
+        """Test retrieval of raw image metadata"""
+        md = self.butler.get("raw_md", visit=self.visit, ccd=self.ccdNum)
+
```
laurenam: Should any specific asserts be made here (e.g. with md.get())?

tests/hscRepository.py
```diff
@@ -0,0 +1,153 @@
...
+import lsst.utils.tests as utilsTests
+from lsst.utils import getPackageDir
+
+TEST_DATA_PACKAGE = "testdata_subaru"
+
+def requirePackage(packageName):
+    """
+    Decorator to skip unit tests unless the package ``packageName`` is set up.
+    """
+    try:
```
laurenam: Is the newline necessary after opening triple quotes (and below)?

bin.src/genCalibRegistry.py
```diff
@@ -17,17 +17,6 @@
 
 import optparse
```
laurenam: Add copyright?

bin.src/registryInfo.py
```diff
@@ -7,17 +7,6 @@
     import sqlite3 as sqlite
 except ImportError:
     import sqlite
```
laurenam: Add copyright?

tests/getRaw.py
```diff
@@ -1,91 +0,0 @@
...
-                    for amp in ccd:
-                        amp = cameraGeom.cast_Amp(amp)
-                        print ccd.getId(), amp.getId(), amp.getDataSec().toString(), \
-                              amp.getBiasSec().toString(), amp.getElectronicParams().getGain()
-                    cameraGeomUtils.showCcd(ccd, ccdImage=raw, frame=frame)
-                    frame += 1
-
-    def testPackageName(self):
-        self.assertEqual(self.butler.mapper.packageName, "obs_subaru")
-
```
laurenam: This test was recently added.  Should it be included in your new test suite?

tests/hscRepository.py
```diff
@@ -0,0 +1,153 @@
...
+    of input data and return its path.
+
+    @param[in] mapperName   Name of mapper class for use with repository.
+    @param[in] inputPath    Location of data files which will be ingested.
+    """
+    repoPath = mkdtemp()
+    with open(os.path.join(repoPath, "_mapper"), "w") as _mapper:
+        _mapper.write(mapperName)
+    ingest_cmd = os.path.join(getPackageDir("obs_subaru"), "bin.src", "hscIngestImages.py")
+    check_call([ingest_cmd, repoPath] + glob(os.path.join(inputPath, "*.fits.gz")))
```
laurenam: I recall @timj mentioning that calls directly to the bin.src directory should specify python explicitly in the command?

timj: Yes, the call should include `sys.executable` explicitly as `bin.src` scripts won't have been fixed. The other option is to use the `bin` version since I have now, in theory, modified the shebang target so that the tests depend on it having executed.

ktlim: I think we should always use the `bin` version, not just because it tests the most but also because it is the most portable (in case we ever manage to get rid of `bin.src` in the future).

timj: Agreed.

bin.src/registryInfo.py
jdswinbank: Good catch.

tests/getRaw.py
jdswinbank: It is, just in a different location: https://github.com/lsst/obs_subaru/pull/11/files#diff-bca4038463bd6bad7345d89f4ccd7618R82

tests/hscRepository.py
jdswinbank: Necessary? No, but I think it's more readable in the code, and it shouldn't make any difference to the processed documentation (per PEP 0257).

tests/hscRepository.py
jdswinbank: Good idea.

tests/hscRepository.py
jdswinbank: If the script is guaranteed to exist in `bin` at the time the test is run, executing that is clearly the right thing to do & I've changed this code appropriately. It'd be helpful if that behaviour/requirement was documented somewhere, though. (Is it?)

tests/hscRepository.py
jdswinbank: I think it is, because the different test cases test different things. It's useful to know if the various components of `IngestRawTestCase` are succeeding even if a problem means that `setUp()` for `GetDataTestCase` fails.

The duplication is minimal, so I'm minded to leave it as it is. If you strongly object, do say so.

tests/hscRepository.py
jdswinbank: Good point; added.

bin.src/genCalibRegistry.py
jdswinbank: Yup.

bin.src/registryInfo.py
jdswinbank: Yup.

tests/hscRepository.py
laurenam: Ok, this does not seem to be working, i.e. the answer to John's query is NO at the moment.  With `testdata_subaru` setup, the `obs_subaru` build fails because the hscRepository.py test fails.  The reason is that the scripts are NOT in the `bin` directory.  I also tried building without `testdata_subaru` setup and the build completes, but again the scripts are NOT being copied to the `bin` directory.  Is there some reason the copy would fail if the bin directory already exists (`obs_subaru` seems to be somewhat unique in that there is already a `bin` directory containing two scripts, `convertFilterData` and `sqlToLog`, which did NOT get moved to the bin.src directory with the rest)?

ktlim: I think this is because of the explicit list of `defaultTargets` in the `SConstruct` file.  I'm not sure why that's there, but if it's needed, "shebang" will have to be added.

jdswinbank: I've been looking at this, and have failed to reproduce @laurenam's error.

I think the `defaultTargets` are explicitly listed in order to add `hsc`, which isn't in the upstream defaults. A better approach might be:
```python
scripts.BasicSConstruct("obs_subaru", defaultTargets=scripts.DEFAULT_TARGETS + ("hsc",))
```
However, since `shebang` is an explicit dependency of `tests`, it seems like it should always be executed before `tests`, regardless of whether it's listed a a default. And, indeed, that's what seems to happen on my system. (Although I have failed to find a way to get scons to tell me explicitly which targets it's building and why -- any suggestions?)

@laurenam, are running with an up-to-date (post DM-4603) `sconsUtils`? I note that currently the default stack on tiger currently has an older version; I'll install a newer version now. I bet that fixes your problem.

laurenam: Confirmed!  All works fine with the newer stack.

PR #14, opened 2016-02-12
Tickets/dm 3373

natelust:


python/lsst/obs/hsc/calibs.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 
 from lsst.pex.config import Field, ListField
-from hsc.pipe.tasks.detrends import FlatCombineConfig, FlatCombineTask
+from lsst.pool.tasks.constructCalibs import FlatCombineConfig, FlatCombineTask
```
natelust: Update this to the proper pipe drivers path

PR #16, opened 2016-03-03
Tickets/dm 4855

SimonKrughoff:


SimonKrughoff:
Please squash the commands that touch the same lines.  I'd even argue that you can squash the first four commits into a single commit on the config.

PR #17, opened 2016-03-14
Tickets/dm 5289

natelust:


config/hsc/processCcd.py
```diff
@@ -19,3 +20,6 @@
 
 # Do not use NO_DATA pixels for fringe subtraction.
 config.isr.fringe.stats.badMaskPlanes=['SAT', 'NO_DATA']
+
+config.charImage.detectAndMeasure.measurement.algorithms["base_Jacobian"].pixelScale = 0.168
+config.calibrate.detectAndMeasure.measurement.algorithms["base_Jacobian"].pixelScale = 0.168
```
natelust: This is not the proper way to configure measurement plugins, see the Jacobian unit test in meas_base

TallJimbo: Is it just "algorithms"->"plugins" you're referring too?  Otherwise I don't see anything obviously wrong here ("algorithms" is present for backwards compatibility, but we do prefer "plugins" now).

PaulPrice: The use of `algorithms` instead of `plugins` appears to be fairly widespread in obs_subaru.  I'll add an extra commit to change them all over.

PR #18, opened 2016-03-15
Review for DM-4831: Add bright object masks to pipeline outputs

jdswinbank:


config/measureCoaddSources.py
```diff
@@ -11,3 +10,15 @@
...
+#
+# This isn't good!  There appears to be no way to configure the base_PixelFlags measurement
+# algorithm based on a configuration parameter; see HSC-1342 for a discussion.  The name
+# BRIGHT_MASK must match assembleCoaddConfig.brightObjectMaskName
+#
+config.measurement.plugins["base_PixelFlags"].masksFpCenter.append("BRIGHT_OBJECT")
+config.measurement.plugins["base_PixelFlags"].masksFpAnywhere.append("BRIGHT_OBJECT")
+
+config.measurement.plugins.names |= ["base_InputCount"]
+
```
jdswinbank: Do we need the extra blank line?

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+import lsst.daf.base as dafBase
+import lsst.pex.logging as pexLog
+import lsst.afw.coord as afwCoord
+import lsst.afw.geom as afwGeom
+import lsst.afw.table as afwTable
+
+class ObjectMaskCatalog(object):
+    """Class to support bright object masks
+
+    N.b. I/O is done by providing a readFits method which fools the (old) butler.
```
jdswinbank: We no longer have a meaningful distinction between "old" and "new" Butler. What "new" Butler feature will break this? Can we refer to a specific ticket?

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    def __init__(self):
+        schema = afwTable.SimpleTable.makeMinimalSchema()
+
+        schema.addField("radius", 'Angle', "radius of mask")
```
jdswinbank: Why do we change from double to single quotes and back? (It doesn't do any harm, it just caught my eye and I'm curious as to why the inconsistency.)

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+import lsst.afw.table as afwTable
+
+class ObjectMaskCatalog(object):
+    """Class to support bright object masks
+
+    N.b. I/O is done by providing a readFits method which fools the (old) butler.
+    """
+    def __init__(self):
+        schema = afwTable.SimpleTable.makeMinimalSchema()
+
```
jdswinbank: (Picky, but...) why is the blank line here? It seems like it should either be removed or be a line lower (so that it appears between ``addField`` and ``self._catalog = ...``).

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+        return self._catalog.__getitem__(i)
+
+    def __setitem__(self, i, v):
+        return self._catalog.__setitem__(i, v)
+
+    @staticmethod
+    def readFits(fileName, hdu=0, flags=0):
+        """Read a ds9 region file, returning a ObjectMaskCatalog object
+
+        This method is called "readFits" to fool the butler. The corresponding mapper entry looks like
```
jdswinbank: Oh I see! Please file a (low-priority) issue for @n8pease and the Data Access team to look into some rainy afternoon.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+
+        with open(fileName) as fd:
+            for lineNo, line in enumerate(fd.readlines(), 1):
+                line = line.rstrip()
+
+                if re.search(r"^\s*#", line):
+                    #
+                    # Parse any line of the form "# key : value" and put them into the metadata.
+                    #
+                    # As noted in HSC-1342, we expect to see at least the keys
```
jdswinbank: Again, please don't refer to HSC tickets.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+
+                line = re.sub(r"^\s*#.*", "", line)
+                if not line:
+                    continue
+
+                if re.search(r"^\s*wcs\s*;\s*fk5\s*$", line, re.IGNORECASE):
+                    checkedWcsIsFk5 = True
+                    continue
+
+                mat = re.search(r"^\s*circle(?:\s+|\s*\(\s*)"
```
jdswinbank: A comment describing exactly what this is supposed to match would be helpful. Indeed, for bonus points, describe in the docstring of this method exactly what the file format you're expecting to read is, providing e.g. links to the DS9 spec where appropriate.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+                                "(\d+(?:\.\d*))([d'\"]*)" "(?:\s*|\s*\)\s*)"
+                                "\s*#\s*ID:\s*(\d+)" "\s*$"
+                                , line)
+                if mat:
+                    ra, raUnit, dec, decUnit, radius, radiusUnit, _id = mat.groups()
+
+                    _id = int(_id)
+                    ra =     convertToAngle(ra,     raUnit,     "ra",     fileName, lineNo)
+                    dec =    convertToAngle(dec,    decUnit,    "dec",    fileName, lineNo)
+                    radius = convertToAngle(radius, radiusUnit, "radius", fileName, lineNo)
```
jdswinbank: Is all that white space necessary? Dangerously close to breaching [the rules](http://developer.lsst.io/en/latest/coding/python_style_guide.html#extraneous-whitespace-should-be-avoided).

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+                    rec["id"] = _id
+                    rec.setCoord(afwCoord.Fk5Coord(ra, dec))
+                    rec["radius"] = radius
+                else:
+                    log.warn("Unexpected line \"%s\" at %s:%d" % (line, fileName, lineNo))
+
+        if not checkedWcsIsFk5:
+            raise RuntimeError("Expected to see a line specifying an fk5 wcs")
+
+        brightObjects._catalog = brightObjects._catalog.copy(True) # allow numpy access to columns
```
jdswinbank: Strictly, the deep copy makes it contiguous in memory so that you can use a {{ColumnView}} which can be exposed to NumPy. I think an extra couple of words here would help make that clear.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+
+        return brightObjects
+
+
+def convertToAngle(var, varUnit, what, fileName, lineNo):
```
jdswinbank: Kind of sad that afw doesn't have this built in.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+    elif varUnit == "'":
+        var /= 60.0
+    elif varUnit == '"':
+        var /= 3600.0
+    else:
+        raise RuntimeError("unsupported unit \"%s\" for %s at %s:%d" %
+                          (radiusUnit, what, fileName, lineNo))
+
+    return var*afwGeom.degrees
+
```
jdswinbank: Do we need the extra blank line?

python/lsst/obs/subaru/objectMasks.py
timj: That is really funny. Hopefully when the Butler solves the FITS bias it will be more obvious how to deal with this.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+        """Read a ds9 region file, returning a ObjectMaskCatalog object
+
+        This method is called "readFits" to fool the butler. The corresponding mapper entry looks like
+        brightObjectMask: {
+            template:      "deepCoadd/BrightObjectMasks/%(tract)d/BrightObjectMask-%(tract)d-%(patch)s-%(filter)s.reg"
+            python:        "lsst.obs.subaru.objectMasks.ObjectMaskCatalog"
+            persistable:   "PurePythonClass"
+            storage:       "FitsCatalogStorage"
+        }
+        and this is the only way I know to get it to read a random file type, in this case a ds9 region file
```
ktlim: There's no inherent FITS bias.  There's just a hack that was added to allow FITS catalogs before a general plugin framework could be developed that happens to provide a way to load arbitrary Python objects.  The principled way to do this currently is (given that a C++ `Storage` subclass is undesirable for a Python object) to add a new Storage to the switch in `daf_persistence:butler.py`.  I think this will actually be a relatively high priority item to enable new types of objects and ways of serializing and persisting them.

python/lsst/obs/subaru/objectMasks.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if varUnit in ("d", ""):
+        pass
+    elif varUnit == "'":
+        var /= 60.0
+    elif varUnit == '"':
+        var /= 3600.0
+    else:
+        raise RuntimeError("unsupported unit \"%s\" for %s at %s:%d" %
+                          (radiusUnit, what, fileName, lineNo))
```
jdswinbank: ``radiusUnit`` is undefined (should be ``varUnit``, I think).

config/measureCoaddSources.py
```diff
@@ -11,3 +10,15 @@
 config.measurement.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsm.py"))
 
 config.deblend.load(os.path.join(getPackageDir("obs_subaru"), "config", "deblend.py"))
+
+config.measurement.plugins["base_PixelFlags"].masksFpAnywhere.append("CLIPPED")
+#
+# This isn't good!  There appears to be no way to configure the base_PixelFlags measurement
+# algorithm based on a configuration parameter; see HSC-1342 for a discussion.  The name
+# BRIGHT_MASK must match assembleCoaddConfig.brightObjectMaskName
```
jdswinbank: Please don't refer to HSC tickets in LSST code.

python/lsst/obs/subaru/objectMasks.py
natelust: This was Robert being cleaver adding comment assuming people would be reading this code after the new butler is already in

PR #21, opened 2016-04-11
DM-3368

r-owen:


config/singleFrameDriver.py
```diff
@@ -0,0 +1,9 @@
+"""
+Subaru-specific overrides for SingleFrameDriverTask (applied before SuprimeCam- and HSC-specific overrides).
+"""
+import os.path
+
+from lsst.utils import getPackageDir
+
+config.processCcd.load(os.path.join(getPackageDir("obs_subaru"), "config", "processCcd.py"))
+config.processCcd.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "processCcd.py"))
```
r-owen: Why is this line present? This looks like a bug to me. It has the effect of always calling HSC-specific overrides even if processing SuprimeCam data, which means that the HSC-specific overrides are being treated as universal for Subaru and should be here, not in the HSC directory.

PaulPrice: Right, there needs to be a `config/hsc/singleFrameDriver.py` to pull the HSC-specific overrides.

natelust: addressed

PR #30, opened 2016-06-30
Review for DM-2674: meas_mosaic

jdswinbank:


config/hsc/processCcd.py
```diff
@@ -13,6 +13,12 @@
 config.charImage.measurePsf.starSelector["objectSize"].fluxMin = 4000
 config.calibrate.refObjLoader.load(os.path.join(hscConfigDir, "filterMap.py"))
 
+# Set to same defaults used in HSC
```
jdswinbank: This *is* HSC!

(Same as on meas_mosaic: in decades to come, that comment will be more confusing than helpful. Maybe better if you can refer to a specific HSC data release this were used for?)

config/Mosaic.py
```diff
@@ -1,3 +1,8 @@
 import os
 from lsst.utils import getPackageDir
 config.astrom.load(os.path.join(getPackageDir("obs_subaru"), "config", "filterMap.py"))
+
+from lsst.pipe.tasks.setConfigFromEups import setPhotocalConfigFromEups
+setPhotocalConfigFromEups(config)
+
+config.load(os.path.join(getPackageDir("obs_subaru"), "config", "srcSchemaMap.py"))
```
jdswinbank: Do we need both `config/Mosaic.py` and `config/hsc/Mosaic.py`? As far as I know, you've only tested on HSC; I'm going to go out on a limb and speculate that meas_mosaic as-is will break if it's applied to SuprimeCam.

Assuming we do need them both, should `srcSchemaMap` live at this level or in `hsc`?

config/hsc/filterMap.py
```diff
@@ -1,2 +1,9 @@
+from lsst.pipe.tasks.setConfigFromEups import setAstrometryConfigFromEups
+menu = { "ps1*": {}, # Defaults are fine
+         "sdss*": {"filterMap": {"y": "z"}}, # No y-band, use z instead
+         "2mass*": {"filterMap": {ff: "J" for ff in "grizy"}}, # No optical bands, use J instead
+        }
+setAstrometryConfigFromEups(config, menu)
```
jdswinbank: Pretty sure you can squash 651625890825 and 510e63560f48 here.

config/processCcd.py
```diff
@@ -31,8 +31,9 @@
                                                 "filterMap.py"))
 
 
-config.calibrate.detectAndMeasure.measurement.plugins['base_ClassificationExtendedness'].fluxRatio = 0.95
 # LAM the following had to be set to affect the fluxRatio used in photoCal in meas_astrom
+# Set to same defaults used in HSC
```
jdswinbank: Same comment as above re HSC.

config/srcSchemaMap.py
```diff
@@ -0,0 +1,209 @@
+config.srcSchemaMap = {
+"deblend_nChild": "deblend_nchild",
```
jdswinbank: I'm surprised by this, as I thought we'd be mapping from HSC field names with `.`s to LSST field names with `_`s, but clearly we're not.

config/srcSchemaMap.py
```diff
@@ -0,0 +1,209 @@
...
+"ext_photometryKron_KronFlux_flag_apCorr": "flux_kron_flag_apcorr",
+"base_GaussianFlux_apCorr": "flux_gaussian_apcorr",
+"base_GaussianFlux_apCorrSigma": "flux_gaussian_apcorr_err",
+"base_GaussianFlux_flag_apCorr": "flux_gaussian_flag_apcorr",
+"calib_psfUsed": "calib_psf_used",
+"calib_psfCandidate": "calib_psf_candidate",
+"calib_psfReserved": "calib_psf_reserved",
+}
+
+#========== No equivalents below =======
```
jdswinbank: What's the difference between these and those marked `no_equivalent` above?

Oh, are these HSC fields which have no equivalent on LSST, whereas the above are LSST fields with no HSC equivalent?

policy/HscMapper.paf
```diff
@@ -398,6 +389,22 @@ datasets: {
         tables:        "raw"
         tables:        "raw_visit"
     }
+    calibrated_src: {
```
jdswinbank: I'm baffled as to why this was previously an exposure!

python/lsst/obs/subaru/isr.py
```diff
@@ -300,10 +297,15 @@ def runDataRef(self, sensorRef):
             flatExposure = self.getIsrExposure(sensorRef, "flat")
             self.flatCorrection(ccdExposure, flatExposure)
 
+        if self.config.doDefect:
+            self.maskAndInterpDefect(ccdExposure, defects)
+        if doWriteDebug: ccdExposure.writeFits(outDir+"/isr7.fits")
```
jdswinbank: Please break the line after the colon.

config/Mosaic.py
laurenam: I've moved things around a bit, but have kept both` Mosaic.py` configs (seeing as we are attempting to maintain the ability to reduce `SuprimeCam` data...if there is a need to adapt `meas_mosaic` to run on such data, this will likely happen as the need is encountered?)  I've pushed a `tickets/DM-2674` branch in `obs_subaru` with all the changes (above and below). Please have a look to see if things are any cleaner in your opinion.

Yes, I moved the `srcSchemaMap` to the `hsc` config level.

config/hsc/filterMap.py
laurenam: Moved the above to the `Mosaic.py` config.

config/hsc/processCcd.py
laurenam: Reworded to clarify.

config/processCcd.py
laurenam: Ditto.

config/srcSchemaMap.py
laurenam: The dots get persisted as underscores in the output schema, so it's really just the name churn that needs accommodating.

config/srcSchemaMap.py
laurenam: Correct...and unnecessary!  I've removed the commented out lines and added a few explanatory notes at the top of the file.

policy/HscMapper.paf
laurenam: !!

python/lsst/obs/subaru/isr.py
laurenam: That actually crept in by mistake, thanks for catching that!  The line has been removed altogether.

jdswinbank: Oh, actually, can we also do:

    os.path.join(outDir, "isr7.fits")

? Thanks!

jdswinbank: Oh, well, good, because I was *also* about to complain about the hard-coded `isr7.fits` filename. :-)

laurenam: Yep, it was just a relic from when I was comparing isr'd exposures at every stage.

config/Mosaic.py
jdswinbank: Thanks -- this is fine.

PR #33, opened 2016-08-25
DM-7378: Updates for Pytest

timj:


tests/testRepository.py
```diff
@@ -64,11 +56,11 @@ def createDataRepository(mapperName, inputPath):
     return repoPath
 
 
-class IngestRawTestCase(utilsTests.TestCase):
+class IngestRawTestCase(lsst.utils.tests.TestCase):
     """Ensure that we can ingest raw data to a repository."""
-    @requirePackage(TEST_DATA_PACKAGE)
+    @unittest.skipIf(testDataPackage is None, "%s not setup"%(testData))
```
timj: I'd prefer to say "is not available" here so as not to imply a specific tool. Also, spacing around the `%`.

tests/testRepository.py
```diff
@@ -21,32 +21,24 @@
...
-    Decorator to skip unit tests unless the package ``packageName`` is set up.
-    """
-    try:
-        getPackageDir(packageName)
-    except pexExcept.NotFoundError as e:
-        return unittest.skip("Required package %s not available [%s]." % (packageName, str(e)))
-    return lambda func: func
+import lsst.utils
 
+testData = "testdata_subaru"
```
timj: This is the package.

tests/testRepository.py
```diff
@@ -21,32 +21,24 @@
...
+try:
+    testDataPackage = lsst.utils.getPackageDir(testData)
```
timj: This is the package directory.

PR #35, opened 2016-09-09
DM-6815: Update LSST full-stack processing configuration to match best practice from HSC

PaulPrice:


config/hsc/processCcd.py
```diff
@@ -19,6 +19,7 @@
 config.calibrate.astrometry.matcher.maxRotationDeg=1.145916
 config.calibrate.astrometry.wcsFitter.numRejIter=3
 
+config.charImage.astrometry.wcsFitter.order = 3
 config.calibrate.astrometry.wcsFitter.order = 3
```
PaulPrice: Are there separate `charImage` and `calibrate` astrometry runs?  If so, we should make sure the settings for both are the same (maybe put them in a loop?).

config/processCcd.py
```diff
@@ -56,12 +56,15 @@
 config.charImage.load(os.path.join(configDir, "cmodel.py"))
 config.charImage.measurement.load(os.path.join(configDir, "apertures.py"))
 config.charImage.measurement.load(os.path.join(configDir, "kron.py"))
+config.charImage.measurement.load(os.path.join(configDir, "hsm.py"))
+if "ext_shapeHSM_HsmShapeRegauss" in config.charImage.measurement.plugins:
+    # no deblending has been done
+    config.charImage.measurement.plugins["ext_shapeHSM_HsmShapeRegauss"].deblendNChild = ""
+    config.charImage.measurement.slots.shape = "base_SdssShape"
```
PaulPrice: If we think HSM is superior, why should we use the SDSS shape here?  I think that if that's what we're doing on HSC, it's by accident.

config/hsc/processCcd.py
RobertLuptonTheGood: You can say
```
config.charImage.astrometry.wcsFitter.order = 3
config.calibrate.astrometry.wcsFitter.order = config.charImage.astrometry.wcsFitter.order
```

PaulPrice: Or:

    for sub in (config.charImage, config.calibrate):
        sub.astrometry.wcsFitter.order = 3

laurenam: Is it always true that these should be the same?  I presumed there may conceivably be some use case for allowing more/less freedom in the characterize vs. calibrated stages.  I'm happy to make the implied desire for using the same value for both here.

config/processCcd.py
laurenam: Actually, looking at your **/tigress/HSC/HSC/rerun/hsc-1382/20151209/cosmos/config/processExposure.py**, HSM measurements are not in the list of algorithms: `processCcd.calibrate.initialMeasurement.algorithms.names=['flux.psf', 'flags.pixel', 'flux.naive', 'flux.gaussian', 'centroid.naive', 'flux.sinc', 'shape.sdss', 'skycoord', 'classification.extendedness']`, so it's not actually being run at this stage.

So, yes, this slot should be changed to `ext_shapeHSM_HsmSourceMoments`.  Now, the question is: do we really want to be running HSM at charImage stage, or just the calibrate stage?  I'm still a bit confused about the mapping from initialMeasurement/calibrate/mesurement (HSC) --> charImage/calibrate (LSST).  It's the initialMeasurement stage on HSC that does not run HSM (the other two do and both have the shape slot of `shape.hsm.moments`).

config/hsc/processCcd.py
PaulPrice: I think all the astrometry settings for both should be the same, so I recommend using a loop over `(config.charImage, config.calibrate)` and modifying the values in the loop.

config/processCcd.py
PaulPrice: `charImage` is the new stage that maps to the `initialMeasurement` in HSC.  I don't think this is important, but I suggest we just allow HSM to sit in the shape slot because we think it's better than SDSS.

PR #37, opened 2016-09-28
Tickets/dm 5503

r-owen:


python/lsst/obs/hsc/makeHscRawVisitInfo.py
```diff
@@ -0,0 +1,73 @@
...
+from lsst.afw.coord import Coord, IcrsCoord, Observatory, Weather
+from lsst.daf.butlerUtils import MakeRawVisitInfo
+
+__all__ = ["MakeHscRawVisitInfo"]
+
+
+class MakeHscRawVisitInfo(MakeRawVisitInfo):
+    """Make a VisitInfo from the FITS header of a Subaru HSC image
+
+    Boresight rotation angle is apparently not available.
```
PaulPrice: What about `INR-STR` and `INR-END`?  I think these are "instrument rotator at start" and "instrument rotator at end".  There's also `INST-PA` which must be "instrument position angle".

python/lsst/obs/hsc/makeHscRawVisitInfo.py
```diff
@@ -0,0 +1,73 @@
...
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import astropy.units
+import astropy.time
```
PaulPrice: Unused import.

python/lsst/obs/hsc/makeHscRawVisitInfo.py
r-owen: Thanks. Do you have specifications for any of these? I would like to know exactly what 0 and 90 degrees mean for the orientation of N and E with respect to detector X and Y (assuming sky rotation, which seems most likely, but a similar question for any other kind of rotation).

r-owen:
I made all suggested changes, rebased and pushed. I documented the possibilities for setting rotator angle but did not implement them yet for lack of necessary information. It is trivial to add this later when the information becomes available.

python/lsst/obs/hsc/makeHscRawVisitInfo.py
PaulPrice: There are no comments on the HSC engineering data I looked at.  Here's what I have from a Suprime-Cam frame (`SUPA01270732.fits`):

    INR-STR =               65.186 / [degree] Inst. rotator angle at start          
    INR-END =               65.185 / [degree] Inst. rotator angle at end            
    INST-PA =               90.000 / Position Angle of Instrument                   

Oooh!  A more recent HSC exposure (`HSCA08609518.fits`) has comments on the headers too:

    INR-STR =              -26.982 / [degree] Inst. rotator angle at start          
    INR-END =              -26.982 / [degree] Inst. rotator angle at end            
    INST-PA =                0.000 / [degree] Position Angle of the Instrument Flang

r-owen: Thanks. The subprime data I have also has that comment; I just don't know what "Position Angle of the Instrument" actually means. I could guess, but it is safer not to!

PR #39, opened 2016-10-18
DM-8013: subtractXTalk is fragile

mrawls:
This makes `subtractXTalk` more robust. Previously, if an exception was raised in the (large) `try` block, the `msk` variable referred to in the `finally` may be undefined. This fix defines `msk` earlier.


mrawls:
This fix looks straightforward and sensible to me - go ahead.

PR #43, opened 2016-11-19
obs_subaru change from paf to yaml mapper policy

n8pease:


pgee2000:
Tested that the yaml file produces the same results as the paf file.  Ordering is now good.

PR #48, opened 2017-01-16
DM-9004: Check uses of darktime for NAN

PaulPrice:


python/lsst/obs/hsc/hscMapper.py
```diff
@@ -192,7 +192,7 @@ def std_raw(self, item, dataId):
 
     def std_dark(self, item, dataId):
         exposure = self._standardizeExposure(self.calibrations['dark'], item, dataId, trimmed=False)
-        visitInfo = afwImage.makeVisitInfo(exposureTime=1.0)
+        visitInfo = afwImage.makeVisitInfo(exposureTime=1.0, darkTime=1.0)
```
r-owen: This looks fine. However, I am curious why darks don't already have correct exposure time (at least), a question that is out of scope, but left over from when I first added VisitInfo support to obs_subaru.

PaulPrice: The dark construction script has never set the dark time.  I'll address that in [DM-8913](https://jira.lsstcorp.org/browse/DM-8913).

PR #50, opened 2017-01-18
DM-6824: Use meas.algorithms.astrometrySourceSelector in measOptimisticB

morriscb:
Added compliance with the new sourceSelector object in meas_astrom/matchOptimisticB as part of tickets/DM-6824. This required a small change to how HSC specific parameters were set in config/hsc/processCdd.py


parejkoj:
This looks ok to me, if it works, but we should check with @PaulPrice for details. Also, please squash those two commits together.

PaulPrice:
Fine with me so long as this goes through ci_hsc.

parejkoj:
Good point: @morriscb please run jenkins on this branch with ci_hsc.

morriscb:
Already done and it does pass ci_hsc.

parejkoj:
Sounds good!

PR #56, opened 2017-02-14
Tickets/dm 9156

pschella:


python/lsst/obs/subaru/SConscript
```diff
@@ -1,5 +1,12 @@
 # -*- python -*-
-from lsst.sconsUtils import scripts
-scripts.BasicSConscript.pybind11(["crosstalk",
-                                  "hscDistortion"])
+from lsst.sconsUtils import env, targets
 
+# Cannot use here?
```
jdswinbank: I dunno if you can use it here or not, but if you can't you don't want it here-but-commented. :-)

ups/obs_subaru.cfg
```diff
@@ -11,5 +11,5 @@ dependencies = {
 config = lsst.sconsUtils.Configuration(
     __file__,
     hasDoxygenInclude=False,
-    hasSwigFiles=True,
+    hasSwigFiles=False,
```
r-owen: I suggest deleting this line (despite the fact that the current default is True)

ups/obs_subaru.table
```diff
@@ -7,7 +7,7 @@ setupRequired(pyfits)
 setupRequired(sconsUtils)
 setupRequired(pybind11)
 
-setupOptional(meas_modelfit)
+#setupOptional(meas_modelfit)
```
r-owen: Please add a short comment explaining why this is commented out and what will be required to restore it (both of which presumably amount to "not yet wrapped with pybind11")

ups/obs_subaru.table
```diff
@@ -7,7 +7,8 @@ setupRequired(pyfits)
 setupRequired(sconsUtils)
 setupRequired(pybind11)
 
-setupOptional(meas_modelfit)
+# temporarily disable meas_modelfit until it is wrapped with pybind11
```
PaulPrice: What's the ticket for re-enabling it?

PR #57, opened 2017-02-14
Tickets/dm 9432

r-owen:


tests/testDistortion.py
```diff
@@ -106,8 +106,8 @@ def getVerificationData():
         CcdData = namedtuple("CcdData", ['name', 'id', 'corners'])
         CornerData = namedtuple("CornerData", ['focalPlane', 'distEst'])
         data = {}
-        for raft in map(afwCG.cast_Raft, camera):
-            for ccd in map(afwCG.cast_Ccd, raft):
+        for raft in map(afwCG.cast_Raft, camera):  # now use: `for raft in camera:`
+            for ccd in map(afwCG.cast_Ccd, raft):  # now use: `for ccd in raft:`
```
kfindeisen: Why are these ones commented instead of being replaced?

r-owen: Because the code in question is part of a comment explaining how a particular dataset was derived using a particular version of the code. The code is still correct for the specified version of the code, but I wanted to show what changes should be made in order to use the code with a modern stack.

kfindeisen: Good grief. Ok, no objections to merging.

PR #58, opened 2017-02-16
isr: mark amp as bad if entirely saturated

PaulPrice:
If the amp is completely saturated, overscan subtraction may fail.

### DIFF
--- a/python/lsst/obs/subaru/isr.py
+++ b/python/lsst/obs/subaru/isr.py
@@ -247,6 +247,11 @@ class SubaruIsrTask(IsrTask):
                 self.saturationDetection(ccdExposure, amp)
             if self.config.doSuspect and not badAmp:
                 self.suspectDetection(ccdExposure, amp)
+            ampMask = afwImage.MaskU(ccdExposure.getMaskedImage().getMask(), amp.getRawDataBBox(),
+                                     afwImage.PARENT)
+            maskVal = ampMask.getPlaneBitMask([self.config.saturatedMaskName, self.config.suspectMaskName])
+            if numpy.all(ampMask.getArray() & maskVal > 0):
+                badAmp = True
             if self.config.doOverscan and not badAmp:
                 ampImage = afwImage.MaskedImageF(ccdExposure.getMaskedImage(), amp.getRawDataBBox(),
                                                  afwImage.PARENT)


python/lsst/obs/subaru/isr.py
```diff
@@ -247,6 +247,11 @@ def runDataRef(self, sensorRef):
                 self.saturationDetection(ccdExposure, amp)
             if self.config.doSuspect and not badAmp:
                 self.suspectDetection(ccdExposure, amp)
+            ampMask = afwImage.MaskU(ccdExposure.getMaskedImage().getMask(), amp.getRawDataBBox(),
+                                     afwImage.PARENT)
+            maskVal = ampMask.getPlaneBitMask([self.config.saturatedMaskName, self.config.suspectMaskName])
+            if numpy.all(ampMask.getArray() & maskVal > 0):
+                badAmp = True
             if self.config.doOverscan and not badAmp:
```
laurenam: I'm not sure I understand the logic here.  maskVal evaluates to 130 (2 for SAT and 128 for SUSPECT), but why are you checking that this evaluates to > 0 (and within the `numpy.all()` call)?

What is the nature of the "failure" in the overscan subtraction?  You are testing for an entirely saturated amp.  Is it true that it is guaranteed not to fail if, say, a single pixel is not saturated (but the rest are)?

PaulPrice: The `> 0` converts the `ampMask.getArray() & maskVal` to a boolean array, which allows use of `numpy.all` to see if all the values are `True`.

I think the overscan subtraction fails because the array it's trying to fit with a spline is empty.

I haven't seen the case of a single unsaturated pixel, but I have seen failures when the amp is entirely saturated.  I suggest we can whack that mole if it pops up.

laurenam: Thanks to @RobertLuptonTheGood for a detailed explanation of the bitwise operator...I now understand what's going on!  He also suggests it may be safer to put parentheses around this operation: `numpy.all((ampMask.getArray() & maskVal) > 0)`

laurenam: And my example of a single unsaturated pixel was an extreme, but I was wondering if it would be better to put a threshold of some sort on calling it a `badAmp` (e.g. if  `> threshold%` of all pixels are saturated and/or suspect, throw in the towel and call it bad).  No need to address that here, but could be something to keep in mind if it ever pops up in practice.

PR #62, opened 2017-03-01
DM-9147: global sky subtraction

PaulPrice:


policy/HscMapper.paf
```diff
@@ -231,6 +231,34 @@ calibrations: {
...
+        storage:     "FitsStorage"
+        level:       None
+    }
+    fringe_camera: {
+        template:    "focalplane/FRINGE/%(calibDate)s/%(filter)s/FRINGE-%(calibDate)s-%(filter)s.fits"
+        persistable: "ExposureF"
+        python:      "lsst.afw.image.ExposureF"
+        storage:     "FitsStorage"
+        level:       None
+    }
```
hsinfang: Why are they here but not in `obs_base`?  They seem useful for non-HSC too.

PR #63, opened 2017-03-01
DM-9501 correct subtask name from astrom to loadAstrom

timothydmorton:


timj:
Closing since the change is no longer needed. Jira ticket was already marked as won't fix.

PR #65, opened 2017-03-10
DM-9111 Add rotator information to HSC VisitInfo

jmeyers314:


python/lsst/obs/hsc/makeHscRawVisitInfo.py
```diff
@@ -67,6 +64,13 @@ def setArgDict(self, md, argDict):
         argDict['era'] = self.eraFromLstAndLongitude(LST, self.observatory.getLongitude())
         argDict['darkTime'] = argDict['exposureTime']
 
+        # Rotation angle formula determined empirically from visual inspection
+        # of HSC images.  See DM-9111.
+        rotAngle = 270.0*degrees - self.popAngle(md, "INST-PA")
+        rotAngle.wrap()
```
r-owen: As of a few days ago `Angle.wrap` now returns a new angle rather than operating in place, so this code requires a minor change. The following should work, or you can break it into two lines if you prefer:
`rotAngle = (270.0*degrees - self.popAngle(md, "INST-PA")).wrap()`

jmeyers314: Ah, great!  I had actually originally tried to use `.wrap()` as you suggest, but I guess this was before the update since it failed at the time.  Change implemented.

PR #74, opened 2017-04-11
DM-9572: Include instrument model in cameraGeom

jmeyers314:


python/lsst/obs/hsc/hscPupil.py
```diff
@@ -0,0 +1,137 @@
...
+
+        @returns horizon rotation angle.
+        """
+        observatory = self.visitInfo.getObservatory()
+        lat = observatory.getLatitude()
+        lon = observatory.getLongitude()
+        radec = self.visitInfo.getBoresightRaDec()
+        ra = radec.getRa()
+        dec = radec.getDec()
+        era = self.visitInfo.getEra()
```
SimonKrughoff: Note that this may not be exactly ERA.  Since it's hard to compute it is typically an approximation.  I don't know if it matters here.

python/lsst/obs/hsc/hscPupil.py
```diff
@@ -0,0 +1,137 @@
...
+        ha = (era + lon - ra).wrap()
+        alt = self.visitInfo.getBoresightAzAlt().getLatitude()
+
+        # parallactic angle
+        sinq = (np.cos(lat.asRadians()) * np.sin(ha.asRadians())
+                / np.cos(alt.asRadians()))
+        cosq = np.sqrt(1 - sinq*sinq)
+        if dec > lat:
+            cosq = -cosq
+        q = Angle(np.arctan2(sinq, cosq))
```
SimonKrughoff: I think the style guide suggests avoiding single character variable names.

tests/testRepository.py
```diff
@@ -147,6 +145,19 @@ def testRaw(self):
         self.assertAlmostEqual(weather.getAirPressure(), self.weath_airPressure)
         self.assertAlmostEqual(weather.getHumidity(), self.weath_humidity)
 
+    def testPupil(self):
+        """Test retrieval of pupil (without checking value)"""
```
SimonKrughoff: Don't you care that it's at least an `HscPupilFactory`?  That seems relatively easy to test.

python/lsst/obs/hsc/hscPupil.py
jmeyers314: Thanks for the notice.  I think approximate is good enough for now, but I'll keep this in mind going forward.

PR #86, opened 2017-08-11
DM-11574: Make testDistortion test the distortion

r-owen:


tests/testDistortion.py
```diff
@@ -22,1238 +22,99 @@
...
+                self.assertPairsAlmostEqual(cornerData.focalPlaneRoundTrip,
+                                            savedCornerData.focalPlaneRoundTrip)
+                self.assertPairsAlmostEqual(cornerData.pixPosRoundTrip, savedCornerData.pixPosRoundTrip)
+
+    def makeDistortionData(self):
+        """Make distortion data
+
+        If SAVE_DATA is true then the data is persisted before the comparison
+        and the test is then reported as failed in order to prevent anyone
+        from committing the test with SAVE_DATA true!
```
PaulPrice: This method doesn't actually do any persistence, so suggest removing this note (which is repeated elsewhere multiple times).

r-owen: Good catch. Thanks!

PR #87, opened 2017-08-11
DM-5922: Rework camera geometry to use the replacement for XYTransform

r-owen:
Update the HSC distortion model to use an astshim PolyMap
Change PUPIL to FIELD_ANGLE
Use lsst.afw.geom.transformRegistry instead of xyTransformRegistry
Update the UPS table to include numpy and astshim


python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -1,7 +1,9 @@
-from lsst.afw.geom import xyTransformRegistry, arcseconds
+import numpy as np
+import astshim as ast
+
+from lsst.afw.geom import transformRegistry, arcseconds, TransformPoint2ToPoint2
 from lsst.pex.config import Config, Field, ListField
-from lsst.obs.subaru import HscDistortion, DistortionPolynomial
-__all__ = ["xyTransformRegistry"]
+__all__ = ["transformRegistry"]
```
PaulPrice: Why do we export something that we've imported? That seems dangerous.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -131,9 +136,11 @@ class HscDistortionConfig(Config):
                                    1.98549941035e-37,
                                    -8.74305862185e-38,
                                    ])
-    skyToCcdOrder = Field(dtype=int, doc="Polynomial order for conversion from sky to CCD (x and y identical)",
+    skyToCcdOrder = Field(dtype=int,
+                          doc="(ignored) Polynomial order for conversion from field angle to focal plane",
```
PaulPrice: If these are ignored, can we just remove them?

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+    maxIter = Field(dtype=int, default=10, doc="Maximum iterations for inversion")  # Usually sufficient
     plateScale = Field(dtype=float, default=1.0, doc="Plate scale (arcsec/mm)")
 
 
+def makeAstPolyMapCoeffs(order, xCoeffs, yCoeffs):
+    """Convert polynomial coefficients in HSC format to AST PolyMap format
+
+    Paramaters
+    ----------
+    order: int
```
PaulPrice: `int`

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+        Polynomial order
+    xCoeffs, yCoeffs: list of float
```
PaulPrice: `list` of `float`

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+        Forward or inverse polynomial coefficients for the x and y axes
+        of output, in this order:
+            x0y0, x0y1, ...x0yN, x1y0, x1y1, ...x1yN-1, ...
+        where N is the polynomial order.
+
+    Returns
+    -------
+    Forward or inverse coefficients for astshim.PolyMap
```
PaulPrice: coeffs : `numpy.ndarray`
        Coefficients for `astshim.PolyMap`.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+        raise ValueError("found %s xCcdToSky params; need %s" % (len(xCoeffs), nCoeffs))
+    if len(yCoeffs) != nCoeffs:
+        raise ValueError("found %s yCcdToSky params; need %s" % (len(yCoeffs), nCoeffs))
+
+    coeffs = np.zeros([nCoeffs * 2, 4])
+    i = 0
+    for nx in range(order + 1):
+        for ny in range(order + 1 - nx):
+            coeffs[i] = [xCoeffs[i], 1, nx, ny]
+            coeffs[i + nCoeffs] = [yCoeffs[i], 2, nx, ny]
```
PaulPrice: What an awful convention for `coeffs`! I suggest that the "HSC format" makes sense, and there should be a helper function in astshim or afw to do the conversion.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+
+    Note that inverse coefficients provided, but they are not accurate enough
+    to use: test_distortion.py reports an error of 2.8 pixels
+    (HSC uses pixels for its focal plane units) when transforming
+    from pupil to focal plane. That explains why the original HSC model uses
+    the inverse coefficients in conjunction with iteration.
+
+    Parameters
+    ----------
+    config: HscDistortionConfig
```
PaulPrice: config : `lsst.obs.subaru.HscDistortionConfig`

It's important to fully-qualify the type because it may be used in a different context.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
+        Distortion coefficients
+
+    Returns
+    -------
+    focalPlaneToPupil: lsst.afw.geom.TransformPoint2ToPoint2
```
PaulPrice: Backticks around type.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -247,20 +255,75 @@ class HscDistortionConfig(Config):
...
     """
-    skyToCcd = DistortionPolynomial(config.skyToCcdOrder, config.skyToCcdOrder,
-                                    config.xSkyToCcd, config.ySkyToCcd)
-    ccdToSky = DistortionPolynomial(config.ccdToSkyOrder, config.ccdToSkyOrder,
-                                    config.xCcdToSky, config.yCcdToSky)
-    return HscDistortion(skyToCcd, ccdToSky, config.plateScale*arcseconds, config.tolerance, config.maxIter)
+    forwardCoeffs = makeAstPolyMapCoeffs(config.ccdToSkyOrder, config.xCcdToSky, config.yCcdToSky)
+
+    # Note that the actual error can be somewhat larger than TolInverse;
+    # the max error I have seen is less than 2, so I scale conservatively
```
PaulPrice: Why is there a need to scale anything? Can't you just use the raw number in the config?

tests/test_distortion.py
```diff
@@ -22,14 +22,15 @@
...
 #
 from __future__ import absolute_import, division, print_function
+import math
 import os.path
 import unittest
 import pickle
 
 import lsst.utils.tests
-from lsst.obs.hsc import HscMapper
+from lsst.obs.hsc import HscMapper, transformRegistry
```
PaulPrice: Shouldn't you pull `transformRegistry` from afw?

tests/test_distortion.py
```diff
@@ -61,8 +65,12 @@ def testDistortion(self):
             self.fail("Cannot find saved data %r; set SAVE_DATA = True and run again to save new data" %
                       dataPath)
 
+        fieldAngleToFocalPlaneTolerance = transformRegistry["hsc"].ConfigClass().tolerance
```
PaulPrice: I think it would be better to put a hard number here rather than using a `Config` value which could change.

python/lsst/obs/hsc/transformRegistry.py
```diff
@@ -1,9 +1,12 @@
+from __future__ import absolute_import, division, print_function
+
+__all__ = ["transformRegistry"]
```
PaulPrice: Squash this part with the previous commit.  (But I don't think we should be exporting `transformRegistry`.)

include/lsst/obs/subaru/HscDistortion.h
```diff
@@ -1,91 +0,0 @@
-// -*- lsst-c++ -*-
```
PaulPrice: Explain in the commit message why it's being removed.

python/lsst/obs/hsc/calibs.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import, division, print_function
```
PaulPrice: Squash with previous commit that had the same goal.

python/lsst/obs/subaru/crosstalkYagi.py
```diff
@@ -214,9 +215,9 @@ def subtractCrosstalkYagi(mi, coeffs1List, coeffs2List, gainsPreampSig, minPixel
        The pixels affected by signal over minPixelToMask have the crosstalkStr bit set
     """
 
-    ####sctrl = afwMath.StatisticsControl()
-    ####sctrl.setAndMask(mi.getMask().getPlaneBitMask("DETECTED"))
-    ####bkgd = afwMath.makeStatistics(mi, afwMath.MEDIAN, sctrl).getValue()
+    # ###sctrl = afwMath.StatisticsControl()
+    # ###sctrl.setAndMask(mi.getMask().getPlaneBitMask("DETECTED"))
+    # ###bkgd = afwMath.makeStatistics(mi, afwMath.MEDIAN, sctrl).getValue()
```
PaulPrice: Just delete these lines.

python/lsst/obs/hsc/transformRegistry.py
r-owen: Because the unit test might fail, as per the comment. I'd rather have the config parameter passed directly to the astshim Mapping, despite it not being a perfect upper limit, rather than try to guess how much to scale it. Guessing here isn't as bad because the test either passes or it doesn't (I added more margin than needed here because the exact amount of error isn't interesting).

tests/test_distortion.py
r-owen: The value I'm testing against is the value used to determine the quantity being tested -- the default value of `tolerance` in the config. Yes if somebody changes that default then the unit test might fail, but that change would have to be done in this package, so the person who changed the default will see the problem right away and be able to deal with it.

python/lsst/obs/hsc/transformRegistry.py
r-owen: OK. As long as you don't mind. In the long run I'd much rather have usable coefficients instead of using an iterative inverse, but that's out of scope for this ticket.

python/lsst/obs/hsc/transformRegistry.py
r-owen: The convention may not be pretty, but it supports N->M axes. We already use 2->2, 2->1 and 1->1 in our stack, so such generality is useful. Let me think about it.

r-owen: I looked into a set of utility functions but decided it was best to leave this as it is for now. This order is reasonable, but swapping x and y axes would be equally valid. I'd rather not assume.

PR #93, opened 2017-09-27
Add mappers for QA parquet tables

timothydmorton:
DM-12030 adds parquet table persistence to the pipe_analysis scripts.  Here, we add the necessary Butler mappers so that they can get the right filenames.

### DIFF
--- a/policy/HscMapper.yaml
+++ b/policy/HscMapper.yaml
@@ -596,4 +596,3 @@ datasets:
     python: __builtin__.str
     storage: TextStorage
     template: plots/%(filter)s/tract-%(tract)d/visit-%(visit)d/%(description)s.parq
-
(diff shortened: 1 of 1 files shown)


timothydmorton:
`filter`, `tract` and `visit` are all camera-agnostic?  Did I hear someone say earlier that `visit` is not, or did I misunderstand?

PaulPrice:
Maybe `visit` is not. But in any case, everything but for the `template` should go in obs_base if you intend that any of this be used for other cameras.

TallJimbo:
`visit` and `ccd` are not camera agnostic.

`tract` and `patch` definitely are.  `filter` is camera-agnostic, but it's values are not.

timothydmorton:
So, I should put `qaTableCoadd` and `qaTableColor` into `obs_base`, leaving `qaTableVisit` in `obs_subaru`?

PaulPrice:
No, everything except the `template` that has `visit` in it goes into obs_base. Then only the `template` that has `visit` in it stays in obs_subaru.

timothydmorton:
OK, so I should include a definition of the `qaTableVisit` in `obs_base`, but just leave `template` blank?  Am I understanding that correctly?

PaulPrice:
Final product looks good. Please squash the three commits before merging.

PR #94, opened 2017-09-28
DM-11976 Fix HSC Policy for python 3 compatibility

sr525:


sr525:
Commit duplicated by DM-12030

PR #96, opened 2017-11-06
DM-12103: Investigate centering for vignetted pupils

jmeyers314:


jmeyers314:
>I'm sure you've thought about this, but just wanted to confirm. With you setting doCenter=True in the init, you're keeping this new functionality as optional, but also making it the default, and thus changing current behaviour. Is this correct? Probably fine as long as you've thought about it.

Yep.  I want it to be the new default.  I seriously doubt anybody but me is currently using this code, so shouldn't be a problem.

PR #97, opened 2017-11-16
DM-12445: Set appropriate default configs for CompareWarp Coadds

yalsayyad:


config/makeCoaddTempExp.py
```diff
@@ -1 +1,4 @@
-config.bgSubtracted = True
+config.makePsfMatched=True
+config.warpAndPsfMatch.psfMatch.kernel['AL'].kernelSize=29
+config.warpAndPsfMatch.psfMatch.kernel['AL'].alardSigGauss=[1.0, 2.0, 4.5]
+config.modelPsf.defaultFwhm=7.7
```
PaulPrice: Spaces either side of the `=` please.

config/coaddDriver.py
```diff
@@ -3,6 +3,12 @@
 
 from lsst.utils import getPackageDir
 
+from lsst.pipe.tasks.assembleCoadd import CompareWarpAssembleCoaddTask
+config.assembleCoadd.retarget(CompareWarpAssembleCoaddTask)
+config.assembleCoadd.assembleStaticSkyModel.subregionSize = (10000, 200)
+config.assembleCoadd.assembleStaticSkyModel.badMaskPlanes = ("BAD", "EDGE", "SAT", "INTRP", "NO_DATA",)
+config.assembleCoadd.assembleStaticSkyModel.doApplyUberCal = True
```
PaulPrice: To eliminate code duplication, this should load the configuration from the configuration file for compareWarpAssembleCoadd (like below). And I think the additional loading below may potentially clobber what you've got here.

yalsayyad: Agreed.  Note that by that reasoning it currently should be reading in safeClipAssembleCoadd.py

laurenam: I think https://jira.lsstcorp.org/browse/DM-10634 might be relevant here.

PR #99, opened 2017-11-28
DM-12765: HSC-specific plugin for recording filter ratios.

TallJimbo:


python/lsst/obs/subaru/filterFraction.py
```diff
@@ -0,0 +1,90 @@
...
+        for ccd in overlapping:
+            filterName = ccd.get(filterKey)
+            counts[filterName] = counts.get(filterName, 0) + 1
+        if "i" in counts:
+            weird = set(counts.keys()) - set(["i", "i2"])
+            if weird:
+                raise lsst.meas.base.FatalAlgorithmError(
+                    "Unexpected filter combination found in coadd: %s" % counts.keys()
+                )
+            measRecord.set(self.key, counts.get("i2", 0.0)/len(overlapping))
```
rearmstr: Do the filter ratios need to take into account the CCD weights?

TallJimbo: They do not.  I had always planned to include that when actually adding the transmission curves, and for some reason I thought it was nonsensical to do it now.  That's clearly not true, and I think I'll have the algorithm produce both a weighted and an unweighted fraction; the weights are already all there.

PR #100, opened 2017-12-07
DM-12947: constructFlat.py fails on HSC

PaulPrice:


python/lsst/obs/hsc/calibs.py
```diff
@@ -77,17 +78,16 @@ def maskVignetting(self, mask, detector):
...
-                x[j, i], y[j, i] = transform.forwardTransform(afwGeom.Point2D(i, j))
-        R = np.hypot(x - self.config.vignette.xCenter, y - self.config.vignette.yCenter)
-        isBad = R > self.config.vignette.radius
+        numPixels = w*h
+        xx, yy = np.meshgrid(np.arange(0, w, dtype=int), np.arange(0, h, dtype=int))
+        xyDetector = [afwGeom.Point2D(x, y) for x, y in zip(xx.reshape(numPixels), yy.reshape(numPixels))]
+        xyFocalPlane = transform.applyForward(xyDetector)
+        origin = afwGeom.Point2D(self.config.vignette.xCenter, self.config.vignette.yCenter)
+        r2 = np.array([pp.distanceSquared(origin) for pp in xyFocalPlane])
+        isBad = (r2 > self.config.vignette.radius**2).reshape((h, w))
```
TallJimbo: What kind of numbers are we looking at for `numPixels` here?  If we're allocating multiple temporaries as big as a CCD, we might want to chunk things up.

TallJimbo: Actually, it looks like we were already allocating temporaries of the same size, so this is nothing new, right?

PaulPrice: `numPixels` is ~8M (2k x 4k). I don't think it's a big deal: we allocate larger temporaries in other places.

PR #101, opened 2017-12-11
tickets/DM-12697

yalsayyad:


config/compareWarpAssembleCoadd.py
```diff
@@ -1,7 +1,8 @@
 config.badMaskPlanes = ("BAD", "EDGE", "SAT", "INTRP", "NO_DATA",)
-config.subregionSize = (10000, 200) # 200 rows (since patch width is typically < 10k pixels
+config.subregionSize = (10000, 200)  # 200 rows (since patch width is typically < 10k pixels
 config.doMaskBrightObjects = True
+config.removeMaskPlanes.append("CROSSTALK")
 config.doNImage = True
-                                                                                                                                                                        
-config.assembleStaticSkyModel.badMaskPlanes = ("BAD", "EDGE", "SAT", "INTRP", "NO_DATA",)
```
pschella: Why is `"NO_DATA"` not kept here? The commit message does seem to imply it is.

config/compareWarpAssembleCoadd.py
```diff
@@ -1,7 +1,8 @@
 config.badMaskPlanes = ("BAD", "EDGE", "SAT", "INTRP", "NO_DATA",)
-config.subregionSize = (10000, 200) # 200 rows (since patch width is typically < 10k pixels
+config.subregionSize = (10000, 200)  # 200 rows (since patch width is typically < 10k pixels
 config.doMaskBrightObjects = True
+config.removeMaskPlanes.append("CROSSTALK")
```
pschella: Slightly confusing on this ticket, but doesn't really matter.

config/compareWarpAssembleCoadd.py
yalsayyad: See change in pipe_tasks. The Non-camera-specific defaults is reset there.

config/compareWarpAssembleCoadd.py
yalsayyad: Was hoping that the separate commit and  message made it clear it is something missed from DM-12665.

PR #103, opened 2017-12-13
DM-12979: add y-band stray light removal

PaulPrice:
LEDs in an encoder in HSC are producing stray light on the detectors,
producing the 'Eye of Y-band' feature. It can be removed by
subtracting open-shutter darks. However, because the pattern of stray
light varies with rotator angle, many dark exposures are required.
To reduce the data volume for the darks, the images have been
compressed using wavelets.

The code here (provided by Sogo Mineo of NAOJ and only cleaned up
a bit) retrieves the appropriate dark, uncompresses it and uses it
to remove the stray light from an exposure.

Some code here (provided by Satoshi Kawanomoto) may not meet coding
standards, but we are accepting it as legacy code. Unfortunately, no
tests are available for this code yet.


python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -0,0 +1,220 @@
...
+        detId = exposure.getDetector().getId()
+        if detId in badAmps:
+            isBad = numpy.zeros_like(model, dtype=bool)
+            for ii in badAmps[detId]:
+                amp = exposure.getDetector()[ii]
+                box = amp.getBBox()
+                isBad[box.getMinY():box.getMaxY(), box.getMinX():box.getMaxX()] = True
+            model[isBad] = numpy.median(model[~isBad])
+            mask = exposure.getMaskedImage().getMask()
+            mask.getArray()[isBad] |= mask.getPlaneBitMask("BAD")
```
TallJimbo: Just aesthetic, but you can now do `mask.array[isBad] |= ...`

python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -0,0 +1,220 @@
...
+
+        model *= exposure.getInfo().getVisitInfo().getExposureTime()
+        array = exposure.getMaskedImage().getImage().getArray()
+        array -= model
```
TallJimbo: The above two lines can also now be written as `exposure.image.array -= model`.

PR #104, opened 2017-12-13
DM-12985: Interpolate before applying B-F correction.

TallJimbo:
Interpolation can only be done on flat-fielded images, so we apply
(and later remove) both that and the dark (for good measure) before
running B-F.

### DIFF
--- a/python/lsst/obs/subaru/isr.py
+++ b/python/lsst/obs/subaru/isr.py
@@ -353,7 +353,7 @@ class SubaruIsrTask(IsrTask):
         if self.config.doSetBadRegions:
             self.setBadRegions(ccdExposure)
 
-        if not interpolationDone:
+        if not interpolationDone or self.config.doNanInterpAfterFlat:
             self.maskAndInterpNan(ccdExposure)
 
         if self.config.qa.doWriteFlattened:
--- a/python/lsst/obs/subaru/isr.py
+++ b/python/lsst/obs/subaru/isr.py
@@ -309,31 +309,21 @@ class SubaruIsrTask(IsrTask):
 
         interpolationDone = False
 
+        darkExposure = self.getIsrExposure(sensorRef, "dark") if self.config.doDark else None
+        flatExposure = self.getIsrExposure(sensorRef, "flat") if self.config.doFlat else None
+
         if self.config.doBrighterFatter:
-
-            # Temporarily apply flat and dark so we can interpolate prior to B-F correction
-            if self.config.doDark:
-                darkExposure = self.getIsrExposure(sensorRef, "dark")
-                self.darkCorrection(ccdExposure, darkExposure)
-            if self.config.doFlat:
-                flatExposure = self.getIsrExposure(sensorRef, "flat")
-                self.flatCorrection(ccdExposure, flatExposure)
-
-            # Interpolate defects, saturation, and NaNs
-            if self.config.doDefect:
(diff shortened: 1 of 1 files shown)


python/lsst/obs/subaru/isr.py
```diff
@@ -295,7 +295,37 @@ def runDataRef(self, sensorRef):
...
+
+            # Interpolate defects, saturation, and NaNs
+            if self.config.doDefect:
+                self.maskAndInterpDefect(ccdExposure, defects)
+            if self.config.doSaturation:
+                self.saturationInterpolation(ccdExposure)
+            self.maskAndInterpNan(ccdExposure)
+            interpolationDone = True
+
+            # Undo flat and dark application
```
PaulPrice: Doing and then undoing suggests using a context manager to simplify the main code path.

python/lsst/obs/subaru/isr.py
```diff
@@ -307,24 +337,23 @@ def runDataRef(self, sensorRef):
             self.darkCorrection(ccdExposure, darkExposure)
         if self.config.doFlat:
             flatExposure = self.getIsrExposure(sensorRef, "flat")
-            self.flatCorrection(ccdExposure, flatExposure)
+            self.flatCorrection(ccdExposure, flatExposure, doTweakFlat=self.config.doTweakFlat)
 
-        if self.config.doDefect:
+        if self.config.doDefect and not interpolationDone:
             self.maskAndInterpDefect(ccdExposure, defects)
```
PaulPrice: Why leave this here?

python/lsst/obs/subaru/isr.py
```diff
@@ -307,24 +337,23 @@ def runDataRef(self, sensorRef):
...
             self.saturationInterpolation(ccdExposure)
 
         if self.config.doFringe:
             self.fringe.runDataRef(ccdExposure, sensorRef)
         if self.config.doSetBadRegions:
             self.setBadRegions(ccdExposure)
 
-        self.maskAndInterpNan(ccdExposure)
+        if not interpolationDone:
+            self.maskAndInterpNan(ccdExposure)
```
PaulPrice: I think we want this to fire always --- there might be zeros in the flat.

RobertLuptonTheGood:
I do not want to test for 0s in the flat;  that's part of the calibs QA.

PaulPrice:
I don't want to test for it either, but I think it's important we can deal with any `NaN`s they cause.

RobertLuptonTheGood:
I'd leave it in as an option but not run it.  It's easy to lose 0.5% here and 1% there.  We could have an option to assert on !NaN

python/lsst/obs/subaru/isr.py
TallJimbo: Where would I move it instead?  I think it's the right place to do defect interpolation if you're not running brighter-fatter (or at least one of several right places, and it's the old one).

PaulPrice: It took me a while to figure out that `not interpolationDone` effectively means `not self.config.doBrighterFatter`. In that case, I think you're right.

python/lsst/obs/subaru/isr.py
TallJimbo: I'm planning to go with @RobertLuptonTheGood's proposal to make this a config option but disable it by default.

PR #107, opened 2018-01-16
DM-13148: Support Suprime-Cam

PaulPrice:


python/lsst/obs/suprimecam/suprimecamMapper.py
```diff
@@ -205,6 +205,9 @@ def defineFilters(self):
     def _extractAmpId(self, dataId):
         return (self._extractDetectorName(dataId), 0, 0)
 
+    def _extractDetectorName(self, dataId):
```
SimonKrughoff: I'm surprised this change in behavior still works.  There still seem to be some things accessed by name.  Is there a different mechanism for getting to the string version of the name (maybe you call `Detector.getName()`)?

config/hsc/ingest.py
```diff
@@ -0,0 +1,8 @@
+from lsst.obs.subaru.ingest import HscParseTask
```
SimonKrughoff: I assume that the user needs to know to pass `--configfile $OBS_SUBARU_DIR/config/hsc/ingest.py`.  Is that true?  If so, can you make sure it's documented and if not, how does the `CmdLineTask` find the configs within a subdirectory?

PaulPrice: Our `ArgumentParser` will load `config/<taskName>.py` and `config/<instrumentName>/<taskName>.py` automatically.

python/lsst/obs/suprimecam/suprimecamMapper.py
PaulPrice: `lsst.obs.base.CameraMapper`'s docstring says:
>     _extractDetectorName(self, dataId): returns the detector name for a CCD
>     (e.g., "CFHT 21", "R:1,2 S:3,4") as used in the AFW CameraGeom class given
>     a dataset identifier referring to that CCD or a subcomponent of it.

It's used to extract from the `dataId` the detector identifier as used in the `camera`.

Suprime-Cam has long had two different means of specifying the detectors: by number and by name (the so-called "Miyazaki names", characters from the anime creator Miyazaki). Since the cameraGeom redesign (I don't think anyone has run this code since then because of the prevalence of HSC), using the names didn't work here: the numbers are required.

timj:
Closing this as no longer relevant. The Jira ticket was closed long ago with Won't Fix.

PR #108, opened 2018-01-26
DM-12375: build and attach TransmissionCurves

TallJimbo:
In addition to support code to make the stuff in obs_base and ip_isr work with HSC, this adds a script to construct TransmissionCurves and install them in a data repository, using data from various sources (mostly the Subaru website).


python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -0,0 +1,153 @@
...
+    wavelengths = wavelengths[i]
+    throughput = throughput[i]
+    if unit == "nm":
+        wavelengths *= 10
+    elif unit != "angstrom":
+        raise ValueError("Invalid wavelength unit")
+    if atMin is None:
+        atMin = throughput[0]
+    if atMax is None:
+        atMax = throughput[-1]
```
PaulPrice: Why pass through `atMin` and `atMax`?

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -0,0 +1,153 @@
...
+
+def getSensorTransmission():
+    """Return a nested dictionary of TransmissionCurves describing the
+    throughput of each sensor.
+
+    Outer directionary keys are string dates (YYYY-MM-DD), with values
+    a dictionary mapping CCD ID to TransmissionCurve.  If the curve
+    is spatially varying, it will be defined in pixel coordinates.
+
+    Dictionary values may be None to indicate that no TransmissionCurve is
```
PaulPrice: Outer dictionary values...

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -0,0 +1,153 @@
...
+def getOpticsTransmission():
+    """Return a dictionary of TransmissionCurves describing the combined
+    throughput of HSC and the Subaru primary mirror.
+
+    Dictionary keys are string dates (YYYY-MM-DD) indicating the beginning of
+    the validity period for the curve stored as the associated dictionary
+    value.  If the curve is spatially varying, it will be defined in focal
+    plane coordinates.
+
+    Dictionary values may be None to indicate that no TransmissionCurve is
```
PaulPrice: `None`
?

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -0,0 +1,153 @@
...
+    a dictionary mapping filter name to TransmissionCurve.  If the curve
+    is spatially varying, it will be defined in pixel coordinates.
+
+    Filter curve names are in the long form used as data ID values (e.g.
+    'HSC-I').
+
+    Dictionary values may be None to indicate that no TransmissionCurve is
+    valid after the date provided in the key.
+    """
+    g = {}
```
PaulPrice: Can you use a better name than `g`? `filterTraces`, maybe?

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -0,0 +1,153 @@
...
+    return short
+
+
+def readTransmissionCurveFromFile(filename, unit="angstrom", atMin=None, atMax=None):
+    wavelengths, throughput = np.loadtxt(os.path.join(DATA_DIR, filename), usecols=[0, 1], unpack=True)
+    i = np.argsort(wavelengths)
+    wavelengths = wavelengths[i]
+    throughput = throughput[i]
+    if unit == "nm":
+        wavelengths *= 10
```
PaulPrice: Why choose Angstroms as the standard? I like Angstroms (it's so wonderfully optical-astronomy), but I think that we've deliberately chosen nm in other places (e.g., `lsst.afw.image.Filter`).

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
+#
+# LSST Data Management System
+#
+# Copyright 2017 AURA/LSST.
```
PaulPrice: 2018

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
+import unittest
+import numpy as np
+
+import lsst.utils.tests
+from lsst.afw.geom import Point2D
+from lsst.obs.hsc import makeTransmissionCurves
```
PaulPrice: I don't see any changes to `python/lsst/obs/hsc/__init__.py` in this changeset. Yet this test must have passed?

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class MakeTransmissionCurvesTest(lsst.utils.tests.TestCase):
+
+    def testOptics(self):
+        wavelengths = np.linspace(4000, 10000, 100)
+        point = Point2D(1000, -500)
+        for curve in makeTransmissionCurves.getOpticsTransmission().values():
```
PaulPrice: Stick result of `makeTransmissionCurves` into `setUp` method?

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+            if curve is None:
+                return
```
PaulPrice: Don't you mean `continue`? Or are you taking advantage of some implicit ordering in the `dict`?

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+            throughputs = curve.sampleAt(point, wavelengths)
+            self.assertTrue((throughputs > 0.70).all())
+
+    def testAtmosphere(self):
+        wavelengths = np.linspace(4000, 12000, 100)
+        point = Point2D(1000, -500)
+        for curve in makeTransmissionCurves.getAtmosphereTransmission().values():
+            if curve is None:
+                return
```
PaulPrice: `continue`?

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+                midR = np.logical_and(wavelengths > 6000, wavelengths < 6300)
+                self.assertTrue((throughputs[midR] > 0.9).all())
+
+    def testFilters(self):
+        wavelengths = np.linspace(3000, 12000, 1000)
+        point = Point2D(1000, -500)
+
+        def check(curve, central, w1, w2):
+            # check that throughput within w1 of central is strictly greater
+            # that throughput outside w2 of central
```
PaulPrice: than

tests/test_makeTransmissionCurves.py
```diff
@@ -0,0 +1,122 @@
...
+                throughput = curve.sampleAt(Point2D(radius, 0), wavelengths)
+                w.append(np.dot(throughput, wavelengths)/throughput.sum())
+            return np.array(w)
+
+        for curves in makeTransmissionCurves.getFilterTransmission().values():
+            r = computeMeanWavelengths(curves['HSC-R'])
+            r2 = computeMeanWavelengths(curves['HSC-R2'])
+            i = computeMeanWavelengths(curves['HSC-I'])
+            i2 = computeMeanWavelengths(curves['HSC-I2'])
+            # i2 and r2 should have lower variance as a function or radius
```
PaulPrice: of radius

tests/test_makeTransmissionCurves.py
TallJimbo: If it's not in `__init__.py`, you can still do the above style of import (or `import lsst.obs.hsc.makeTransmissionCurves`).  You could not do, say,
```
import lsst.obs.hsc as obsHsc
d = obsHsc.makeTransmissionCurves.getOpticsTransmission()
```
Given that this module isn't needed by the vast majority of the code that uses obs_subaru, I figured I'd leave it out of the default imports.  Happy to change that if you think users will find it surprising; I don't think it's a big deal either way.

PaulPrice: Of course --- you're importing the module itself. Duh, silly me.

python/lsst/obs/hsc/makeTransmissionCurves.py
TallJimbo: I didn't really have a strong opinion originall, but angstroms is what I RFC'd back when I started this project, and the afw `TransmissionCurve` classes that document Angstroms as the standard have already been merged.  So I'd rather not backtrack on that now.

python/lsst/obs/hsc/makeTransmissionCurves.py
TallJimbo: Frequently the tabulated values (for the filters in particular) get down close to zero on the edges and would clearly extrapolate to zero, but don't quite get all the way there.  I think it's probably best to set those to exactly zero outside the tabulated area, and this provides a way to do that.

python/lsst/obs/hsc/makeTransmissionCurves.py
TallJimbo: I'll call it `module`, since it's a dict of the module scope of the `.py` file.

tests/test_makeTransmissionCurves.py
TallJimbo: It's a different result for every test method, so I don't think it's worthwhile.

tests/test_makeTransmissionCurves.py
TallJimbo: I thought I was just making use of the fact that these dicts have only one or two elements, but now that I think about it that's bad, because I don't know the first element is the not-`None` one.  Will fix.

PR #109, opened 2018-01-29
DM-13260: Support construction of new HSC calibs

PaulPrice:


config/assembleCoadd.py
```diff
@@ -3,6 +3,7 @@
 config.doMaskBrightObjects = True
 config.removeMaskPlanes.append("CROSSTALK")
 config.doNImage = True
+config.badMaskPlanes += ["SUSPECT"]
```
rearmstr: Is there anything else we currently flag as SUSPECT?  We will now be masking those, maybe we should have also been doing it before.

PaulPrice: `SUPSECT` is also used to mask pixels that are beyond linearity but not saturated. I don't think that there are actually such pixels in practise in HSC, but they should always be masked.

PR #112, opened 2018-02-05
DM-12796 Cleanup jointcal dataset template

parejkoj:


parejkoj:
@PaulPrice requested to not change the template, though I'm pretty sure photoCalib isn't used (measMosaic does its own thing).

laurenam:
I agree with @parejkoj that this change shouldn’t cause problems for `meas_mosaic`’s “native” **wcs** and **fcr** outputs, which are defined here: https://github.com/lsst/obs_subaru/blob/master/policy/HscMapper.yaml#L58

`meas_mosaic` does also persist a PhotoCalib, and I don’t see any reason not to put it in the default/standard location (which just adds the filter subdirectory)...unless the PhotoCalib from `meas_mosaic` is being used by something I’m unaware of?

PR #115, opened 2018-02-09
Tickets/dm 12700

yalsayyad:


config/coaddDriver.py
```diff
@@ -3,6 +3,12 @@
 
 from lsst.utils import getPackageDir
 
+from lsst.pipe.tasks.assembleCoadd import CompareWarpAssembleCoaddTask
+config.assembleCoadd.retarget(CompareWarpAssembleCoaddTask)
+
+# configs specific to CompareWarpAssembleCoaddTask
+config.assembleCoadd.assembleStaticSkyModel.subregionSize = (10000, 200)
```
TallJimbo: Why is this being set here?  I wouldn't have thought there was anything special about the `CoaddDriver` + `CompareWarp` *combination*.

yalsayyad: Because it's laoding the configs from config/assembleCoadd.py instead of config/compareWarpAssembleCoadd.py. 

As lines 6-9 are written, changing "assembleCoadd" -->  "compareWarpAssembleCoadd" doesn't work, because it will then try to do: `config.compareWarpAssembleCoadd.<configs>` which of course isn't the name of the subtask.  (this was actually that last thing that I changed back after I posted the review).   Actually, a better solution is to loop over tuples of the subtask name and config file name. I'll do that.

PR #117, opened 2018-02-21
DM-11138: add jointcal_wcs and jointcal_photoCalib datasets

TallJimbo:


policy/HscMapper.yaml
```diff
@@ -642,3 +642,7 @@ datasets:
   qaTableVisit:
     template: plots/%(filter)s/tract-%(tract)d/visit-%(visit)d/%(description)s.parq
 
+  jointcal_wcs:  # astrometric calibration produced by jointcal/meas_mosaic
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_wcs-%(visit)07d-%(ccd)03d.fits
+  photoCalib:  # photometric calibration produced by jointcal/meas_mosaic
```
laurenam: Missing the "jointcal_" prefix to **photoCalib** here.

policy/HscMapper.yaml
```diff
@@ -642,3 +642,7 @@ datasets:
...
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_photoCalib-%(visit)07d-%(ccd)03d.fits
```
laurenam: Why not put these up at line ~356 along with the deprecated **photoCalib**?  That way the reader sees the newly favored dataset right away.

policy/SuprimecamMapper.yaml
```diff
@@ -497,3 +497,8 @@ datasets:
     python: hsc.pipe.tasks.multiband.MultiBandConfig
     storage: ConfigStorage
     template: config/multiband.py
+
+  jointcal_wcs:  # astrometric calibration produced by jointcal/meas_mosaic
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_wcs-%(visit)07d-%(ccd)03d.fits
+  jointcal_photoCalib:  # photometric calibration produced by jointcal/meas_mosaic
+    template: jointcal-results/%(filter)s/%(tract)04d/jointcal_photoCalib-%(visit)07d-%(ccd)03d.fits
```
laurenam: Don't need the extra space, and I think it makes more sense to include these just before the coadd definitions (i.e. at line ~312...but I'm not insisting!)

PR #124, opened 2018-04-05
DM-14008: Enable transmission curve coaddition for HSC by default.

TallJimbo:
Should have been done long ago, but was overlooked somehow.


TallJimbo:
Thanks for the fix, @PaulPrice - this is better than what I would have come up with, even if I had time to debug.

PR #125, opened 2018-04-26
DM-12635: Gen3 Butler specializations

TallJimbo:
Note that this PR is relative to the new `gen3-middleware` integration branch, not master.

### DIFF
--- a/python/lsst/obs/subaru/gen3.py
+++ b/python/lsst/obs/subaru/gen3.py
@@ -25,8 +25,15 @@
 import re
 
 from lsst.daf.butler.instrument import Instrument
+from lsst.daf.butler.gen2convert import Translator, KeyHandler, ConstantKeyHandler, CopyKeyHandler
 
 __all__ = ("HyperSuprimeCam",)
+
+
+# Regular expression that matches HSC PhysicalFilter names (the same as Gen2
+# filternames), with a group that can be lowercased to yield the
+# associated AbstractFilter.
+FILTER_REGEX = re.compile(r"HSC\-([GRIZY])2?")
 
 
 class HyperSuprimeCam(Instrument):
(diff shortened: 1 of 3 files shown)


python/lsst/obs/subaru/gen3.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, mapper):
+        self.sensors = [
+            dict(
+                sensor=sensor.getId(),
+                name=sensor.getName(),
+                # getType() returns a pybind11-wrapped enum, which
+                # unfortunately has no way to extract the name of just
+                # the value (it's always prefixed by the enum type name).
+                purpose=str(sensor.getType()).split(".")[-1],
```
pschella: Have you tried adding `py::arithmetic()` to the `enum` wrapper? That gives you conversion to value too I think.

python/lsst/obs/subaru/gen3.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+
+class HscAbstractFilterKeyHandler(KeyHandler):
+    """KeyHandler for HSC filter keys that should be mapped to AbstractFilters.
+    """
+
+    __slots__ = ()
+
+    REGEX = re.compile(r"HSC\-([GRIZY])2?")
```
pschella: Why compile it here but not above?

python/lsst/obs/subaru/gen3.py
TallJimbo: I'll open another ticket for that; I'd rather not have to make afw changes for this one.

python/lsst/obs/subaru/gen3.py
TallJimbo: Heh, just wrote them at completely different times.  Will fix.

python/lsst/obs/subaru/gen3.py
TallJimbo: https://jira.lsstcorp.org/browse/DM-14247

PR #131, opened 2018-05-14
Fix gen2->gen3 translation for HSC filter data ID keys.

TallJimbo:


python/lsst/obs/subaru/gen3.py
```diff
@@ -95,42 +95,37 @@ def extract(self, gen2id, skyMap, skyMapName):
...
 Translator.addRule(ConstantKeyHandler("camera", "Camera", "HSC"),
                    camera="HSC", gen2keys=("ccd",), consume=False)
 
-# Copy Gen2 'visit' to Gen3 'exposure' for raw only.
+# Copy Gen2 'visit' to Gen3 'exposure' for raw only.  Also consume filter,
+# since that's implied by 'exposure' in Gen3.
 Translator.addRule(CopyKeyHandler("exposure", "Exposure", "visit"),
-                   camera="HSC", datasetTypeName="raw", gen2keys=("visit",))
+                   camera="HSC", datasetTypeName="raw", gen2keys=("visit",),
+                   consume=("visit", "filter"))
```
pschella: What happens with `filter`? Shouldn't this go to `abstract_filter` or `physical_filter`?

TallJimbo: It turns out we don't have any datasets that use both Visit and AbstractFilter, and Visit + PhysicalFilter is redundant, because a Visit knows its own PhysicalFilter (or, in other words, Visit's dependency on PhysicalFilter is optional).  So stripping `filter` whenever we see `visit` keeps us from adding an incorrect "required" dependency on PhysicalFilter.

We do translate `filter` to `abstract_filter` when its paired with `tract` (via a later rule), and there's at least one DatasetType-specific rule for the one case where we have a PhysicalFilter that isn't from a Visit or Exposure.

PR #136, opened 2018-07-06
DM-7886: pyfits to astropy

timj:


bin.src/showPsfs.py
```diff
@@ -249,8 +249,8 @@ def main(dataDir, visit, title="", outputTxtFileName=None,
     plt.axis([-20000, 20000, -20000, 20000])
 
     def frameInfoFrom(filepath):
-        import pyfits
-        with pyfits.open(filepath) as hdul:
+        from astropy.io import fits
```
r-owen: Please consider moving this import to the top of the file. It's more standard and easier to find there. Also I think `main` always calls `frameInfoFrom`.

PR #137, opened 2018-07-10
DM-15045: Activate convolved measurements on undeblended sources

PaulPrice:


TallJimbo:
Actually, don't you need some new dataset definitions in the mapper policy file, too?

PaulPrice:
Only for the HSC afterburner, so it's on [the HSC branch for obs_subaru](https://github.com/lsst/obs_subaru/commit/51d780df0c14a0ddffcfb9c7cb84971775e757d8).

PR #138, opened 2018-07-13
DM-15105: Style guide fixes

timj:
* Enable flake8 and travis


timj:
@PaulPrice this checking has uncovered a seeming bug:
```
./python/lsst/obs/subaru/ccdTesting.py:2844:92: F821 undefined name 'median'
./python/lsst/obs/subaru/ccdTesting.py:2847:14: F821 undefined name 'medianFilterImage'
```

Looking at `arcturus.py` it seems there is an undisclosed dependency on `lsst.analysis` and that is where `medianFilterImage` should be coming from. I don't see `setupRequired(analysis)` in the table file. `median` does not seem to be defined anywhere so I'm not sure how you want me to fix this.

python/lsst/obs/subaru/ccdTesting.py
```diff
@@ -1002,7 +986,7 @@ def grads(exp):
     plt.clf()
     for a in ccd:
         ima = im[a.getRawDataBBox()].getArray()
-        I = np.mean(ima, 1)/np.median(ima)
+        I = np.mean(ima, 1)/np.median(ima)  # noqa E741
```
timj: @PaulPrice not sure if you would rather rename this variable to fix the warning or keep the noqa.

timj:
And matplotlib is imported but not declared in the table file, and `import lsst.afw.display.ds9 as ds9` is being used.

bin.src/showPsfs.py
```diff
@@ -356,26 +362,26 @@ def frameInfoFrom(filepath):
...
 
-        plt = main(dataDir, visit, title, args.outputTxtFile,
-                   gridPoints=args.gridPoints,
-                   showFwhm=args.showFwhm, minFwhm=args.minFwhm, maxFwhm=args.maxFwhm,
-                   correctDistortion=args.correctDistortion,
-                   showEllipticity=args.showEllipticity, ellipticityDirection=args.ellipticityDirection,
-                   showNdataFwhm=args.showNdataFwhm, showNdataEll=args.showNdataEll,
-                   minNdata=args.minNdata, maxNdata=args.maxNdata,
-                   verbose=args.verbose)
+        plot = main(dataDir, visit, title, args.outputTxtFile,
```
timj: This one was causing trouble because `plt` was clashing with the `plt` that was imported from `matplotlib`.

PaulPrice:
Let's jettison ccdTesting.py. OK, @RobertLuptonTheGood ?

RobertLuptonTheGood:
OK.  But there's potentially useful stuff in there, so I'd like to see any files removed from obs_subaru put onto a branch (called something like HSCcommissioningTools) before deleting them

timj:
I've discussed this on Slack and the plan is to remove the file and be explicit in the commit message. I'm having trouble pushing this change to the branch at the moment.

timj:
@PaulPrice I've made the changes, and added a few more cleanups? Can you take a look or should I ask @r-owen ?

bin.src/genDefectRegistry.py
```diff
@@ -67,11 +65,12 @@ def __init__(self, path, version, validStart, validEnd=None):
 # Defects files for a CCD are valid from the date they are registered until the next date.
 # This means that any defects file should carry ALL the defects that are present at that time.
 for ccd, rowList in rowsPerCcd.items():
-    rowList.sort(key=lambda row: row.validStart) # ISO-8601 will sort just fine without conversion from str
+    # ISO-8601 will sort just fine without conversion from srt
```
r-owen: type: "srt" -> "str"

bin.src/makeGeom.py
```diff
@@ -236,7 +234,7 @@ def printElectronics(buff, ccdId, ccdname, xidx, yidx, infoA, infoB):
 
     # 0th layer is pure metadata
     for ccd in range(1, len(ptr)):
-        #for ccd in [1, 28]:
+        # for ccd in [1, 28]:
```
r-owen: Consider deleting commented-out code. This looks even safer to remove than the bit earlier.

bin.src/makeGeom.py
```diff
@@ -146,8 +143,8 @@ def printRaftCcdGeom(buff, ccdId, ccdname, xidx, yidx, xpos, ypos):
             buff.write('    serial: -1 \n')
 
     buff.write('    Ccd: { \n')
-    #buff.write('        name: "R:0,0 S:%d,%d CFHTid:%d %s" \n' % (xidx, yidx, ccdId, ccdname))
-    #buff.write('        serial: -1 \n')
+    # buff.write('        name: "R:0,0 S:%d,%d CFHTid:%d %s" \n' % (xidx, yidx, ccdId, ccdname))
+    # buff.write('        serial: -1 \n')
```
r-owen: Consider deleting commented-out code. I can see why you kept it -- better safe than sorry, especially as it's not your package. @PaulPrice what do you think?

setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+ignore = E133, E226, E228, E251, N802, N803, N806
+exclude = __init__.py, config/*.py, suprimecam/* hsc/*, bin/*
+
+[tool:pytest]
+addopts = --flake8
+flake8-ignore = E133 E226 E228 E251 N802 N803 N806
+    config/*.py ALL
+    suprimecam/* ALL
+    hsc/* ALL
+    bin/* ALL
```
r-owen: Is the `ALL` necessary? I have not seen it used before.
Also, why ignore everything in `bin`? I would expect all the python code to be from `bin.src` and to pass.

ups/obs_subaru.build
```diff
@@ -1,2 +0,0 @@
-@LSST BUILD@ &&
-build_lsst @PRODUCT@ @VERSION@ @REPOVERSION@
```
r-owen: Why is this file going away?

python/lsst/obs/subaru/strayLight/waveletCompression.py
```diff
@@ -14,24 +14,22 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-#----------------------------------------------------------------------------
+# ----------------------------------------------------------------------------
```
r-owen: I know this is picky and feel free to ignore it, but consider getting rid of the line?

ups/obs_subaru.build
timj: There was a conversation on Slack last week where we all came to the conclusion that nothing uses these files and they are unused cruft in the system that confuse people.

python/lsst/obs/subaru/strayLight/waveletCompression.py
timj: Yes. Sorry. I should have removed it.

setup.cfg
timj: Testing `bin/` and `bin.src/` doesn't add anything so I thought it was quicker to ignore them. The `ALL` might not be necessary, I didn't think it was until last week when a package I was working on with @SimonKrughoff required it. I'll take another look.

bin.src/genDefectRegistry.py
timj: I have no idea how I transposed those letters.

setup.cfg
timj: I have experimented. See also https://pypi.org/project/pytest-flake8/

Specifying a file in the `flake8-ignore` section without any parameter does nothing. You need the `ALL` to prevent pytest from looking at the file at all. flake8 does respect the exclude line in the section above though. So without the ALL pytest sends the file to flake8 and flake8 returns a PASS because it was told to excliude the file. If you specify `ALL` then pytest never sends the file to flake8 so it doesn't even look like a test. `ALL` is therefore the correct approach.

timj:
I have made the requested changes. I'll wait until tomorrow before merging in case @PaulPrice has a comment.

laurenam:
Just FYI, @PaulPrice is away through Wednesday.

bin.src/makeGeom.py
PaulPrice: I think this file is left over from the cameraGeom transition half a lifetime ago, and it looks like it contains CFHT-specific stuff (?!). Suggest we simply dump it all.

bin.src/refileFocus.py
```diff
@@ -12,8 +12,6 @@
     108		110
```
PaulPrice: I don't know what this script does. Suggest dumping it.

ups/obs_subaru.build
PaulPrice: That is correct.

PR #143, opened 2018-07-24
DM-12315: Generalize ap_pipe to non-HiTS data

kfindeisen:
This PR adds HSC-specific settings for `ApPipeTask` and `AssociationTask` to obs package and camera override files.


config/hsc/apPipe.py
```diff
@@ -0,0 +1,9 @@
+# Config override for lsst.ap.pipe.ApPipeTask
+import os.path
+from lsst.utils import getPackageDir
+
+hscConfigDir = os.path.join(getPackageDir('obs_subaru'), 'config', 'hsc')
+
+config.ccdProcessor.load(os.path.join(hscConfigDir, "processCcd.py"))
```
mrawls: It seems like this line is intended to supersede line 7 in `config/apPipe.py` above, because it's specific to the HSC camera rather then the Subaru telescope in general. Are you sure it will take precedence?

kfindeisen: Yes, everything in a camera-specific config file supersedes anything in an observatory config file. This is done in [other config files](https://github.com/lsst/obs_subaru/blob/master/config/hsc/singleFrameDriver.py#L8), too.

PR #144, opened 2018-08-01
DM-15244: Change fluxSigma to fluxErr and similarly for apCorr and covariances

r-owen:


sr525:
I'm not sure that this is related to the fluxSigma to fluxErr changes.

PR #148, opened 2018-08-22
DM-15189: Initial implementation of Gen3 raw-data ingest

TallJimbo:
Strongly recommend looking at each of the two commits here separately; one is almost entirely just moving code around.

### DIFF
--- a/tests/test_ingest.py
+++ b/tests/test_ingest.py
@@ -1,0 +1,179 @@
+# This file is part of obs_subaru.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


tests/test_ingest.py
```diff
@@ -0,0 +1,191 @@
...
+try:
+    testDataDirectory = lsst.utils.getPackageDir(testDataPackage)
+except lsst.pex.exceptions.NotFoundError as err:
+    testDataDirectory = None
+
+
+TESTDIR = os.path.dirname(__file__)
+
+
+skipIfNoData = unittest.skipIf(testDataDirectory is None, "testdata_subaru must be set up")
```
timj: Since none of the tests run if data are not available it might be less repetitive to put the skip in `setUpClass` or `setUp`.

TallJimbo: Ah, I didn't realize the skip decorators worked on classes, too.  Will do.

timj: I haven't looked deeply but skipping at the class level vs during `setUp` *might* result in different skip statistics. I don't remember if there is a difference.

PR #163, opened 2018-11-04
DM-15073: draft Object.yaml

yalsayyad:
I'm not attached to the file's location.  I am amenable to moving it deeper or to qa_explorer for now

### DIFF
--- a/policy/Object.yaml
+++ b/policy/Object.yaml
@@ -1,0 +1,392 @@
+funcs:
+    objectId: # the index of deepCoadd_obj IS the objectId
+        functor: Index
+    parentObjectId:
+        functor: Column
+        args: parent
+        dataset: ref
+    coord_ra:
+        # reference position for merged "ref" cat. Required by db. Not in DPDD
+        functor: RAColumn
+        dataset: ref
+    coord_dec:
+        # reference position for merged "ref" cat. Required by db. Not in DPDD
+        # coord_dec because "dec" is reserved in most SQL DBs
+        functor: DecColumn
+        dataset: ref
+    Ra:
+        functor: RAColumn
+        dataset: meas
+    Dec:
+        functor: DecColumn
+        dataset: meas
+    # raErr:  not available yet DM-15180
+    # decErr: not available yet DM-15180
+    refBand: # reference band is same for all measurements
+        functor: ReferenceBand
+        dataset: ref
+    # psRadecTai: No moving point source model yet
+    # psRadec: No moving point source model yet
+    # psPm: No moving point source model yet
+    # psParallax: No moving point source model yet
+    PsFlux:
+        functor: NanoJansky
+        dataset: forced_src  # recommend meas too
+        args: slot_PsfFlux_instFlux
+    PsFluxErr:
+        functor: NanoJanskyErr
+        dataset: forced_src
+        args:
+            - slot_PsfFlux_instFlux
+            - slot_PsfFlux_instFluxErr
+    # Assume the database will provide UDFs for Mags
+    # psCov:
+    #     Flux should be somewhere in the name?
+    #     How best to store cov matrix?
+    # psLnL:  No moving point source model yet
+    # psChi2:  No moving point source model yet
+    # psNdata: No moving point source model yet
+
+    # DRP is working the the Bulge-Disk model fluxes
+    # In the meantime, CModel is the best
+    # CModel fluxes have alias slot_ModelFlux
+    # bdRadec: # Same as SdssCentroid
+    BdE1:
+        functor: e1
+        dataset: meas
+        args:
+            - modelfit_CModel_ellipse_xx
+            - modelfit_CModel_ellipse_xy
+            - modelfit_CModel_ellipse_yy
+    # rBdE1Err:  Not computed
+    BdE2:
+        functor: e2
+        dataset: meas
+        args:
+            - modelfit_CModel_ellipse_xx
+            - modelfit_CModel_ellipse_xy
+            - modelfit_CModel_ellipse_yy
+    # rBdE2Err:  Not computed
+    BdReB:
+        functor: RadiusFromQuadrupole
+        dataset: meas
+        args:
+            - modelfit_CModel_dev_ellipse_xx
+            - modelfit_CModel_dev_ellipse_xy
+            - modelfit_CModel_dev_ellipse_yy
+    BdReD:
+        functor: RadiusFromQuadrupole
+        dataset: meas
+        args:
+            - modelfit_CModel_exp_ellipse_xx
+            - modelfit_CModel_exp_ellipse_xy
+            - modelfit_CModel_exp_ellipse_yy
+    #bdCov: Not computed
+    BdChi2:
+        functor: Column
+        dataset: meas
+        args: modelfit_CModel_objective
+    #bdSamples: Not computed
+    BdFluxB:
+        functor: NanoJansky
+        dataset: meas
+        args: modelfit_CModel_dev_instFlux
+    BdFluxBErr:
+        functor: NanoJanskyErr
+        dataset: meas
+        args:
+            - modelfit_CModel_dev_instFlux
+            - modelfit_CModel_dev_instFluxErr
+    BdFluxD:
+        functor: NanoJansky
+        dataset: meas
+        args: modelfit_CModel_exp_instFlux
+    BdFluxDErr:
+        functor: NanoJanskyErr
+        dataset: meas
+        args:
+            - modelfit_CModel_exp_instFlux
+            - modelfit_CModel_exp_instFluxErr
+    grStd:
+        # HSC users still debating
+        #  1) PSF convolved undeblended aperture fluxes?
+        #  2) Scarlet outputs?
+        # In the meantime: forced CModel
+        functor: Color
+        dataset: forced_src
+        args: slot_ModelFlux
+        filt1: HSC-G
+        filt2: HSC-R
+    riStd:
+        functor: Color
+        dataset: forced_src
+        args: slot_ModelFlux
+        filt1: HSC-R
+        filt2: HSC-I
+    izStd:
+        functor: Color
+        dataset: forced_src
+        args: slot_ModelFlux
+        filt1: HSC-I
+        filt2: HSC-Z
+    zyStd:
+        functor: Color
+        dataset: forced_src
+        args: slot_ModelFlux
+        filt1: HSC-Z
+        filt2: HSC-Y
+    refIxx:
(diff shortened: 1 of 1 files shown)


policy/Object.yaml
```diff
@@ -0,0 +1,381 @@
...
+        functor: ReferenceBand
+        dataset: ref
+    # psRadecTai: No moving point source model yet.
+    # psRadec: No moving point source model yet.
+    # psPm: No moving point source model yet.
+    # psParallax: No moving point source model yet.
+    PsFlux:
+        functor: NanoJansky
+        filt: HSC-R
+        # Should we include forced_src too?
```
TallJimbo: If we have to pick one, I think including *just* the `forced_src` flux would be more appropriate; the model described by the DPDD uses the same position (and motion) across all bands.

policy/Object.yaml
```diff
@@ -0,0 +1,381 @@
...
+    #     How best to store cov matrix?
+    # psLnL:  No moving point source model yet.
+    # psChi2:  No moving point source model yet.
+    refPsNdata:
+        # This does not make sense outside a moving point source model,
+        # because slot_PsfFlux_area is just the footprint on the coadd and depends on seeing
+        # At the moment it only makes sense to do this per band
+        # DPDD thinks this is an integer. It is not.
+        functor: Column
+        args: slot_PsfFlux_area
```
TallJimbo: What the DPDD wants isn't just the footprint area, because the pixels we use are defined by the size of the PSF model bounding box, not the footprint.  In any case, the larger point that it's not very meaningful when we're not doing a moving PS is certainly true.

However, `base_PsfFlux_area` isn't really related to any of these (and that's why it's not an integer); it's a measure of PSF width that has units of area rather than units of length (and sadly, "effective area" means something totally different from "effective radius").  That's a useful quantity, but not one in the DPDD at all.

policy/Object.yaml
```diff
@@ -0,0 +1,381 @@
...
+            - modelfit_CModel_dev_ellipse_yy
+    BdReD:
+        functor: EffectiveRadius
+        dataset: meas
+        args:
+            - modelfit_CModel_exp_ellipse_xx
+            - modelfit_CModel_exp_ellipse_xy
+            - modelfit_CModel_exp_ellipse_yy
+    #bdCov: Not computed
+    #bdChi2: Not computed
```
TallJimbo: `modelfit_CModel_objective` is the chi^2.

TallJimbo:
In terms of location for this file, I don't think it needs to be hidden away in qa_explorer, but I also don't think it's at all HSC specific; it only seems that way because HSC data is what we regularly process through the prototype DRP pipelines, and I didn't see anything here I would expect to change if we were applying these functors to, say, DESC DC2 images.  I think obs_base, pipe_tasks, or a new package (presumably with other parquet-producing tasks) would be a better home.

policy/Object.yaml
yalsayyad: Is what the DPDD wants something that lets us go from Chi^2 to reduced Chi^2? I'm going to call `psNdata` "Not computed yet,"

TallJimbo: Yes, that's my understanding.

yalsayyad:
There are explicit declarations of "HSC-R", "HSC-G" etc in here. We should be able to remove all references to filter names in the future and make the functors loop over filters even for colors in a smart way. 

For now, I don't want the letters "HSC" in that order to appear anywhere in `obs_base`, but I'll keep this in mind. After the meeting I'll copy over some DC2 data to lsst-dev we can play with.

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
+funcs:
+    objectId: # the index of deepCoadd_obj IS the objectId
+        functor: Index
+    parentObjectId:
+        functor: Column
+        args: parent
+        dataset: ref
+    coord_ra:
+        # reference position for merged "ref" cat. Required by db. Not in DPDD
+        functor: RAColumn
```
laurenam: Should the “A” of RA really be capitalized?

Also, there’s a note about why coord_dec below, but no accompanying one for coord_ra here.  Deliberate?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        dataset: ref
+    coord_dec:
+        # reference position for merged "ref" cat. Required by db. Not in DPDD
+        # coord_dec because "dec" is reserved in most SQL DBs
+        functor: DecColumn
+        dataset: ref
+    Ra:
+        functor: RAColumn
+        dataset: meas
```
laurenam: Ditto re “A”

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        functor: DecColumn
+        dataset: meas
+    # raErr:  not available yet DM-15180
+    # decErr: not available yet DM-15180
+    refBand: # reference band is same for all measurements
+        functor: ReferenceBand
+        dataset: ref
+    # psRadecTai: No moving point source model yet.
+    # psRadec: No moving point source model yet.
+    # psPm: No moving point source model yet.
```
laurenam: Should the “d”s of dec be capitalized here?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+    #     Flux should be somewhere in the name?
+    #     How best to store cov matrix?
+    # psLnL:  No moving point source model yet.
+    # psChi2:  No moving point source model yet.
+    # psNdata: No moving point source model yet.
+
+    # DRP is working the the Bulge-Disk model fluxes
+    # In the meantime, CModel is the best
+    # CModel fluxes have alias slot_ModelFlux
+    #bdRadec: # Same as SdssCentroid
```
laurenam: Ditto re “d”

Space needed after #

Here and throughout, inconsistent use of period on comments.

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        functor: Column
+        args: slot_Shape_yy
+        dataset: meas
+    Ixy:
+        functor: Column
+        args: slot_Shape_xy
+        dataset: meas
+    # Icov: # need to compute
+    # DPDD should include Psf Shape:
+    refIxxPSF:
```
laurenam: “PSF” -> “Psf”? (and many cases below)

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+    # petroRad50Err:
+    # petroRad90:
+    # petroRad90Err:
+    KronRad:
+        # Convert to sky coords
+        functor: Column
+        args: ext_photometryKron_KronFlux_radius
+        dataset: meas  # or forced_src?
+    # kronRadErr: # Not computed
+    # kronBand: replaced with `refBand`
```
laurenam: “k” vs. “K”?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        args: ext_photometryKron_KronFlux_instFlux
+    KronFluxErr:
+        functor: NanoJanskyErr
+        dataset: meas  # or forced_src?
+        args:
+            - ext_photometryKron_KronFlux_instFlux
+            - ext_photometryKron_KronFlux_instFluxErr
+    # kronRad50: Need to compute DM-16313
+    # kronRad50Err: Need to compute DM-16313
+    #  kronRad90: Need to compute DM-16313
```
laurenam: Extra space after #

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        dataset: ref
+    xy_flag:
+        functor: Column
+        args: slot_Centroid_flag
+        dataset: ref
+    # blendedness: 'base_Blendedness_abs_flux'
+    FWHM:
+        functor: Seeing
+        dataset: meas
+    refFWHM:
```
laurenam: Should “FWHM” really be all capitals?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+    - ext_photometryKron_KronFlux_flag_used_psf_radius
+    - slot_PsfShape_flag
+    - slot_PsfShape_flag_no_pixels
+    - slot_PsfShape_flag_not_contained
+    - slot_PsfShape_flag_parent_source
+    - slot_Shape_flag
+    - slot_Shape_flag_no_pixels
+    - slot_Shape_flag_not_contained
+    - slot_Shape_flag_parent_source
+# Need to add per object Galactic extinction to post-processing
```
laurenam: Sims is now installed with the shared stack, so could this be added using functions there (as is done in `pipe_analysis`). It’s not part of `lsst_distrib`, but then neither is `qa_explorer`, yet... I’m not sure where the dependency boundaries lie...

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+        dataset: meas
+        args:
+            - modelfit_CModel_exp_instFlux
+            - modelfit_CModel_exp_instFluxErr
+    grStd:
+        # HSC users still debating
+        #  1) PSF convolved undeblended aperture fluxes?
+        #  2) Scarlet outputs?
+        # In the meantime: forced CModel
+        functor: Color
```
laurenam: Is this a Std from a given locus?  Is the function to compute that really just called “Color”?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+            - modelfit_CModel_dev_ellipse_yy
+    BdReD:
+        functor: RadiusFromQuadrupole
+        dataset: meas
+        args:
+            - modelfit_CModel_exp_ellipse_xx
+            - modelfit_CModel_exp_ellipse_xy
+            - modelfit_CModel_exp_ellipse_yy
+    #bdCov: Not computed
+    BdChi2:
```
laurenam: “bd” vs “Bd”?

policy/Object.yaml
```diff
@@ -0,0 +1,392 @@
...
+    Extendedness:
+        functor: Column
+        args: base_ClassificationExtendedness_value
+        dataset: meas
+    # lcPeriodic:  # Not computed. Need multi-epoch forced phot S19
+    # lcNonPeriodic:  # Not computed. Need without multi-epoch forced phot S19
+    # photoZ:  # Not computed yet
+
+    # COLUMNS ADDED (not in the DPDD)
+    x:
```
laurenam: Is the unit/frame here pixel/tract?

policy/Object.yaml
yalsayyad: I don't a problem. `RAColumn` is the name of Tim's Functo,r and the column names turn into `gRa`, `rRa`, `iRa`as specified by cat and the dpdd.

PR #174, opened 2019-01-14
DM-17195: Fix missing INST-PA message for y-band data.

czwa:
The rewrite of ISR to be more general also attempted to remove as many
extra butler calls as possible.  However, as ci_hsc does not test
y-band data, this caused a hidden bug with the y-straylight
correction.  The required keywords should not be pulled from the
exposure header, but rather from the VisitInfo associated with the
exposure.  The MJD-STR and MJD-END keywords are still pulled from the
header, so this may need to be revisited if those keywords are moved
into the VisitInfo as well.

### DIFF
--- a/python/lsst/obs/subaru/strayLight/rotatorAngle.py
+++ b/python/lsst/obs/subaru/strayLight/rotatorAngle.py
@@ -20,6 +20,7 @@ __all__ = ["inrStartEnd"]
 
 import numpy as np
 from astropy.io import fits
+import lsst.geom
 
 # fixed parameters
 ltt_d = 19.82556  # dome latitude in degree
@@ -232,13 +233,13 @@ def _minorArc(angle1, angle2):
     return angle1, angle2
 
 
-def inrStartEnd(header):
+def inrStartEnd(visitInfo):
     """Calculate instrument rotator angle for start and end of exposure
 
     Parameters
     ----------
-    header : `lsst.daf.base.PropertySet`
-        FITS header for exposure to correct
+    visitInfo : `lsst.afw.image.VisitInfo`
+        Visit info for the exposure to calculate correction.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


python/lsst/obs/subaru/strayLight/rotatorAngle.py
```diff
@@ -248,9 +249,14 @@ def inrStartEnd(header):
         Instrument rotator angle at end of exposure, degrees.
     """
 
-    inst_pa = header.getDouble('INST-PA')
-    ra_t_d = header.getDouble('CRVAL1')
-    de_t_d = header.getDouble('CRVAL2')
+    visitInfo = exposure.getInfo().getVisitInfo()
+    header = exposure.getMetadata()
+
+    inst_pa = 270.0 - visitInfo.getBoresightRotAngle().asAngularUnits(lsst.geom.degrees)
```
timj: The 270 has already been applied to getBoresightRotAngle(). See https://github.com/lsst/astro_metadata_translator/blob/master/python/astro_metadata_translator/translators/hsc.py#L132-L138

python/lsst/obs/subaru/strayLight/rotatorAngle.py
```diff
@@ -248,9 +249,14 @@ def inrStartEnd(header):
...
+    ra_t_sp, de_t_sp = visitInfo.getBoresightRaDec()
+
+    ra_t_d = ra_t_sp.asAngularUnits(lsst.geom.degrees)
+    de_t_d = de_t_sp.asAngularUnits(lsst.geom.degrees)
 
     mjd_str = header.getDouble('MJD-STR')
```
timj: Use `visitInfo.getDate()`.  This returns the middle of the exposure. Then use `visitInfo.getExposureTime()` to get the exposure time and add and subtract half of that. `ObservationInfo` does include explicit start and end times but obs packages do not yet have direct access to that.

python/lsst/obs/subaru/strayLight/rotatorAngle.py
```diff
@@ -248,9 +249,14 @@ def inrStartEnd(header):
...
     """
 
-    inst_pa = header.getDouble('INST-PA')
-    ra_t_d = header.getDouble('CRVAL1')
-    de_t_d = header.getDouble('CRVAL2')
+    visitInfo = exposure.getInfo().getVisitInfo()
+    header = exposure.getMetadata()
+
+    inst_pa = 270.0 - visitInfo.getBoresightRotAngle().asAngularUnits(lsst.geom.degrees)
+    ra_t_sp, de_t_sp = visitInfo.getBoresightRaDec()
```
timj: I had not idea this would work. Now I need to find out how `getBoresightRaDec()` can return a `SpherePoint` object or ra and dec separately.

python/lsst/obs/subaru/strayLight/rotatorAngle.py
```diff
@@ -232,13 +233,13 @@ def _minorArc(angle1, angle2):
     return angle1, angle2
 
 
-def inrStartEnd(header):
+def inrStartEnd(exposure):
```
timj: It might just need a VisitInfo.

RobertLuptonTheGood:
The rule is that *nothing* ever comes from headers/metadata;  anything we need must be captured as instance variables/objects.  So I agree with Tim; MJD delenda est.

python/lsst/obs/subaru/strayLight/rotatorAngle.py
czwa: I saw that, but as the straylight code expects the header value of INST-PA, my assumption is that I need to put it back.  I did some tests comparing against the results from before the bug, and the residuals are best with the 270 included.

python/lsst/obs/subaru/strayLight/rotatorAngle.py
czwa: I also wasn't sure it would work, but it looked likely based on the results I had in the debugger.

PR #177, opened 2019-02-04
DM-17552: Check SQuaSH metric regressions using Match PessimisticB

morriscb:


morriscb:
Hey Paul, it seems as if the config/processCcd.py setting doesn't completely do the job. Using the validate_drp scripts (even after modifying the configs there) default grabs the astrometry.net catalogs instead of the newer ones. Changing the config in config/hsc resolves the issue and the script loads the correct refcats. Is there some other way to resolve the issue? Should we edit the hsc config but not the base config?

PaulPrice:
Sounds like there's a bug in validate_drp (maybe [here](https://github.com/lsst/validate_drp/blob/master/config/hscConfig.py#L13-L15).
For this package, editing the one under `config` is the right thing to do, because we want this setting to be the  same for all cameras in this package.

morriscb:
As I mentioned in the last comment, editing this config does not fix the problem. Take a look at the pull request [here](https://github.com/lsst/validate_drp/pull/99/files). Despite this change I still also needed this change to obs_subaru to get things working.

PaulPrice:
It has to be a bug in how validate_drp is getting run, because we regularly process HSC data with the PS1 refcat without this change.

RobertLuptonTheGood:
It's possible that `--show history=...` would help sort this out.

					R

morriscb:
Okay, after working with @jdswinbank for a bit it looks like it was a stale config sitting around. The changes to validate_drp and meas_astrom are the only pull requests needed. Thanks for having me look into this more you two. I'll close the pull request and delete the branch.

PR #185, opened 2019-03-07
DM-18174: fix calib construction

PaulPrice:


config/makeBrighterFatterKernel.py
```diff
@@ -1,7 +1,5 @@
 from lsst.obs.subaru.crosstalk import CrosstalkTask
-from lsst.obs.subaru.isr import SubaruIsrTask
 
-config.isr.retarget(SubaruIsrTask)
 config.isr.crosstalk.retarget(CrosstalkTask)
```
czwa: This should import and retarget `SubaruCrosstalkTask` from `lsst.obs.subaru.crosstalk`.

RobertLuptonTheGood: Why do we need any Subaru-specific tasks?  I thought @czwa had removed the need for them

czwa: The ISR crosstalk correction and the Subaru crosstalk correction have different algorithms, so I didn't put them together.  I have thought about swapping things (ISR.crosstalk -> obs_decam.crosstalk; obs_subaru.crosstalk -> ISR.crosstalk), but couldn't come up with a clear justification.
CT needs to be updated to work with Gen3 butler, so this will need to be revisited anyway.

RobertLuptonTheGood: So the problem is that the BF kernel code didn't get updated.

PR #187, opened 2019-04-04
DM-17830: Use only stars in internal jointcal fitting

yalsayyad:


config/hsc/jointcal.py
```diff
@@ -7,3 +7,16 @@
...
+config.sourceSelector['science'].doUnresolved = True
+config.sourceSelector['science'].doSignalToNoise = True
+config.sourceSelector['science'].signalToNoise.minimum = 10.
+config.sourceSelector['science'].doIsolated = False
+config.sourceSelector['science'].doFlags = True
+config.sourceSelector['science'].doFluxLimit = True
+config.sourceSelector['science'].fluxLimit.minimum = 10
+badFlags = ['base_PixelFlags_flag_edge', 'base_PixelFlags_flag_saturated',
+            'base_PixelFlags_flag_interpolatedCenter', 'base_SdssCentroid_flag',
+            'base_PsfFlux_flag', 'base_PixelFlags_flag_suspectCenter']
```
parejkoj: Jointcal also uses `slot_Shape_flag` in its badFlags list. I don't know whether that is redundant with the others or not.

config/hsc/jointcal.py
```diff
@@ -7,3 +7,16 @@
...
 config.colorterms.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "colorterms.py"))
+
+config.sourceSelector.name = 'science'
+config.sourceSelector['science'].doUnresolved = True
+config.sourceSelector['science'].doSignalToNoise = True
+config.sourceSelector['science'].signalToNoise.minimum = 10.
+config.sourceSelector['science'].doIsolated = False
+config.sourceSelector['science'].doFlags = True
+config.sourceSelector['science'].doFluxLimit = True
+config.sourceSelector['science'].fluxLimit.minimum = 10
```
parejkoj: Why do you have both `fluxLimit` and `signalToNoise` cuts? I don't see what a minimum flux gains over a SN cut.

Also, I think `fluxLimit` should be named `instFluxLimit`, since that's what it's actually cutting on.

config/hsc/jointcal.py
```diff
@@ -7,3 +7,16 @@
 config.applyColorTerms = True
 
 config.colorterms.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "colorterms.py"))
+
+config.sourceSelector.name = 'science'
+config.sourceSelector['science'].doUnresolved = True
+config.sourceSelector['science'].doSignalToNoise = True
+config.sourceSelector['science'].signalToNoise.minimum = 10.
+config.sourceSelector['science'].doIsolated = False
```
parejkoj: Why not cut to just isolated sources? I'm not sure I'd trust blended sources for either astrometry or photometry.

config/hsc/jointcal.py
yalsayyad: That's true only if the user is using the `astrometry` source selector`!  I am not!  The name difference should make this more obvious.  (`flags.bad` not `badFlags`)

Your `JointcalConfig.setDefaults` is highly nonstandard btw.  We the select default source selectors by doing `self.sourceSelector.name = astrometry`. `astrometry` is now "active."  If your desired config change applies to all source selectors in the registry, you can access the "active" source selector's config via `self.sourceSelector.active.`  If that's not guaranteed, I specify which one it applies to `self.sourceSelector['astrometry'].badFlags.extend(["slot_Shape_flag"])`.  For example if I were to do `self.sourceSelector.active.badFlags.extend(["slot_Shape_flag"])` in the file here,  it'd fail because `science` doesn't have a `badFlags`

yalsayyad:
That's a good idea to add some inline comments regarding the choices. I'll add those.

config/hsc/jointcal.py
yalsayyad: Hmm. That's worth a rerun. I'll try that out.

config/hsc/jointcal.py
yalsayyad: Because `signalToNoise` is based on PsfFlux (configurable), and negative `CalibFluxes` were being included and were crashing jointcal.  Remember that?

config/hsc/jointcal.py
RobertLuptonTheGood: So the plan is to set the default for `doIsolated` to `True`?  I don't think I'd remove the option

config/hsc/jointcal.py
parejkoj: What do you mean by "highly nonstandard" in this context? Can you show me the documentation on what a `config.setDefaults` is supposed to look like?

My comment above was just that you have a bad flags list here that did not include `slot_Shape_flag`, and I was wondering if it should be in your list or no.t

config/hsc/jointcal.py
parejkoj: I only vaguely remember it. If that's the reason, we definitely need to document why this is set here. And I wonder if a minimum flux even closer to zero is even more appropriate? Is `1` a better choice than `10`?

parejkoj: I just had a thought related to this: if we're using PSF sources now, you should probably also set `config.sourceFluxType="Psf"` so that the `PsfFlux` fields are used instead of `CalibFlux`, no?

config/hsc/jointcal.py
yalsayyad: Sure. Your use of `setDefaults` is fine, it is the setting of the  *RegistryField* that is weird:

https://pipelines.lsst.io/modules/lsst.pipe.base/command-line-task-retargeting-howto.html#how-to-retarget-a-subtask-configured-as-a-registryfield-with-a-configuration-file

yalsayyad: Sorry, I did misunderstand your comment about missing the `slot_Shape_flag`.  I assumed it was necessary if you were including galaxies, but not if you were selecting bright point sources.

config/hsc/jointcal.py
yalsayyad: I'll kick off a rerun as an experiment, but I don't know enough about calibration to assess this.  I've only ever seen aperture fluxes of stars used for calibration.  Paging @erykoff

erykoff: I have been paged!  And I'm confused...I don't know why psf sources are involved in this at all... but to be clear:
a) `slot_CalibFlux` is an alias.  It is by default set to `base_CircularApertureFlux_12_0`, which is what we want to calibrate to.  But if somebody wants to calibrate something else (psf flux?  model flux?), then they can change this.  The photometric calibration code should simply use `calibFlux` and trust the system.
b) If the default for `signalToNoise` is set to `PsfFlux`, this is incorrect if we're using this selection for calibration.  It should be overridden to `CalibFlux` and then we're not mixing in the psf modeling (which is undesirable).  And then if you change the calib flux field, this selection will change with it.
c) Having the `fluxLimit` as well is odd.  Why 10?  Or 1?  Depends on the units, which is strange (but is probably fine for HSC, we know the units, this isn't general).  But if we fix the s/n cut flux field (and we should!), then (I believe) this will become unnecessary.
d) I see John's point about this being `instFluxLimit` in this case, but in general it's just a flux limit and you can set the flux field to anything you want, inst or not.  But again, this is outside the scope of this ticket.

RobertLuptonTheGood: You usually use aperture fluxes as they are less subject to systematic noises than PSF fluxes (e.g. brighter fatter).  Because we are calibrating (using many brightish stars) the lower noise of the PSF flux isn't important.

config/hsc/jointcal.py
yalsayyad: @RobertLuptonTheGood I don't follow which direction 
> I don't think I'd remove the option
is going.  You leaning True or False?

yalsayyad:
^^^ github is lying. I just made this branch the same as master.

PR #202, opened 2019-05-21
DM-19214: ip_isr crosstalk shouldn't depend on detector hasCrosstalk( )and getCrosstalk()

czwa:
With ip_isr now supporting configuration defined crosstalk correction, there's no need for obs_subaru/HSC to use a separate implementation.  Switch HSC crosstalk to use ISR implementation.


timj:
@czwa what's holding up this pull request?

PR #205, opened 2019-05-30
DM-18739: Ingest defects via new standard format.

SimonKrughoff:
As a part of this, defects have been moved to [obs_subaru_data](https://github.com/lsst/obs_subaru_data).


policy/HscMapper.yaml
```diff
@@ -182,6 +181,22 @@ exposures:
...
+    python: lsst.meas.algorithms.Defects
+    refCols:
+    - visit
+    reference: raw_visit
+    storage: FitsStorage
+    tables: defects
+    template: defects/%(calibDate)s/defects-%(calibDate)s-%(ccd)03d.fits
+    validEndName: validEnd
+    validRange: true
+    validStartName: validStart
```
czwa: Although the defects are generated and registered as before, the new files do not appear to be automatically picked up as they are on master.  For a test of an arbitrary (non-`ci_hsc`) file:
```
runIsr.py /datasets/hsc/repo --rerun private/czw/DM-18739/tagv --id visit=36234 ccd=24 --clobber-versions --clobber-config
```
I receive a "sqlite3.OperationalError: no such table: defects", I assume because /datasets/hsc/repo/ does not have these updated files.  Appending the new CALIB directory to my path (`--rerun $OBS_SUBARU_DIR/hsc/CALIB/:private/czw/DM-18739/tagv`) gives a error that the two input repos do not have the same (or common) parents.  If there's an obvious solution I'm missing, I'd like to test it to ensure that this change doesn't break processing outside of tests like `ci_hsc`.

If the solution is that all repos need to have the new defects added by copying/merging the generated version from `$OBS_SUBARU_DIR/hsc/CALIB/`, that seems like it should be documented.

SimonKrughoff: I will make sure that this is documented when we make the cut over.  It should be easy to do.  I don't think there is a way to have two different calib repositories with the gen2 system.

czwa: As long as it's announced so it's not a surprise, and all the existing calib repos get updated, this should be fine.

PR #206, opened 2019-06-07
Add dataset type for fakes QA plots

sr525:


policy/HscMapper.yaml
```diff
@@ -650,6 +650,13 @@ datasets:
     python: builtins.str
     storage: TextStorage
     template: plots/%(filter)s/tract-%(tract)d/visit-%(visit)d/plot-v%(visit)d-%(description)s-%(style)s.png
+  plotFakes:
+    description: "QA plots using fakes, made by fakesAnalysis as part of visitAnalysis."
+    python: None
+    storage: MatplotlibStorage
+    tables: raw
+    persistable: None
+    template: plots/%(filter)s/tract-%(tract)d/visit-%(visit)d/fakesPlots/plot-v%(visit)d-%(description)s.png
```
mrawls: I have no issue with adding this to a mapper in a general sense, but shouldn't the same policy definitions be available for other non-hsc cameras too?

PR #216, opened 2019-07-26
DM-20763: Add initial support for Gen3 Butler to obs_decam

parejkoj:


python/lsst/obs/hsc/hscMapper.py
```diff
@@ -84,15 +81,9 @@ def __init__(self, **kwargs):
 
         self.addFilters()
 
-        #
-        # self.filters is used elsewhere, and for now we'll set it
-        #
-        # It's a bit hard to initialise self.filters properly until #2113 is resolved,
```
timj: Do we know whether this comment is still relevant?

python/lsst/obs/subaru/gen3/hsc/rawFormatter.py
```diff
@@ -39,25 +39,14 @@ class HyperSuprimeCamRawFormatter(FitsRawFormatterBase):
...
 
     def makeWcs(self):
         wcs = makeSkyWcs(self.metadata, strip=True)
         dimensions = bboxFromMetadata(self.metadata).getDimensions()
         center = Point2D(dimensions/2.0)
         return makeFlippedWcs(wcs, self.FLIP_LR, self.FLIP_TB, center)
 
-    def makeFilter(self):
-        # For historical reasons we need to return a short, lowercase filter
-        # name that is neither a physical_filter nor an abstract_filter in Gen3
```
timj: This comment is quite confusing because it seems it always used physical_filter for lookup.

parejkoj: Yes, this is very confusing to me too. I think I've identified the issue that @TallJimbo was dealing with when he wrote this. We should talk about it today to decide on an approach.

python/lsst/obs/hsc/hscMapper.py
parejkoj: #2113 is from trac: https://dev.lsstcorp.org/trac/ticket/2113

We now have `getCanonicalName()`, as used below. But I think we want to rethink Filter sooner rather than later, as the behavior of `getName()` with aliases is very confusing to me.

python/lsst/obs/subaru/gen3/hsc/instrument.py
```diff
@@ -79,18 +87,8 @@ def register(self, registry):
                 # HSC doesn't have rafts
                 raft=None
```
timj: Oh, HSC does have rafts as well. `1` and `2` and `ObservationInfo` knows about them. @TallJimbo why can't we add them?

TallJimbo: I don't mind if we do.  I'm not sure we have a use case, but I guess there's a chance it might be useful, and not declaring any rafts doesn't have any chance of being useful.

PaulPrice: What is meant by "raft" these days? My understanding is that the HSC focal plane consists of a single Invar surface, but the CCDs are divided into two electronic backends. So does "raft" refer to a mechanical or electronic separation?

timj: In astro_metadata_translator I call them "detector groups". THere is code in obs_subaru bin.src that uses the term raft so I assumed obs_subaru would like to have access to the "raft/detector group" in gen 3.

tests/test_ingest.py
```diff
@@ -39,126 +36,15 @@
...
-            self.runIngest(file=[self.file, "nonexistent.fits"])
-        try:
-            dt = self.butler.registry.getDatasetType("raw.metadata")
-        except KeyError:
-            # If we also rollback registering the DatasetType, that's fine,
-            # but not required.
-            pass
-        else:
-            self.assertIsNotNone(self.butler.registry.find(self.butler.collection, dt, self.dataId))
+        super().setUp()
```
TallJimbo: This is a bit of a nitpick, but seeing this in action makes me think that it might be slightly better for the base class interface to be an `abstractmethod` that takes all of the `setUp` attributes as arguments, which would be called by the base-class `setUp`.  That makes it harder for the derived class to forget something and get delayed, harder-to-interpret errors.  Of course, this is just test code, so it's not like it's a case where immediate vs. delayed errors are a big deal.

timj: Also, since the setUp being called here is not the unittest setUp, it might be clearer to rename the method in the parent class so it's clear you are calling the IngestTestBase one and not the TestCase one.

python/lsst/obs/subaru/gen3/hsc/rawFormatter.py
parejkoj: The new `FilterDefinition` system should be much more explicit about the various names, so I think we've cleared this up.

python/lsst/obs/subaru/gen3/hsc/instrument.py
parejkoj: Can we deal with that on a separate ticket, please?

tests/test_ingest.py
parejkoj: @timj : I don't understand what you mean by "not the unittest setUp"? It's the MRO's job to sort that out. And `unittest` doesn't do anything in `setUp` anyway.

timj: Why leave it to the MRO? It seems to me to be much clearer to have no doubt as to which parent method is being called. Calling it `setUp` gives it connotations that `TestCase` is involved somewhere.

parejkoj: I'm still confused: this *is* `setUp`, which is only called if this is a `TestCase`. That's the whole point. The MRO will call whatever parent `setUp`s are available, in the appropriate order.

timj: Does python really call all `setUp`s that are available? Doesn't it only do that if the method explicitly asks the parent to be called? If you think it's not at all confusing to people which `setUp` is being called (but I don't think it's both) then leave it. I'm saying I have to think about which one is being called and if it was called something else then that would be clearer to me (one of the parent classes is not a TestCase).

parejkoj: `TestCase` does not define `setUp` though.

TallJimbo: I am less concerned with the naming here and more with the fact that you are passing objects to the base `setUp` by setting them as attributes before the call, rather than just passing them naturally as arguments.

python/lsst/obs/subaru/gen3/hsc/instrument.py
parejkoj: Similarly to my comment on the obs_decam PR, the HSC things are not `rafts`, they are groups of detectors. Any such definition of a grouping of detectors should be dealt with on another ticket.

PR #219, opened 2019-08-21
DM-20548: Update HSC's camera geometry to use units of millimeters

laurenam:


bin.src/showCamera.py
```diff
@@ -25,26 +25,44 @@
...
 
 import lsst.afw.cameraGeom as cameraGeom
+import lsst.afw.geom as afwGeom
 import lsst.geom as geom
 
+debug = False
+
 
 def main(camera, sample=20, names=False, showDistortion=True, plot=True, outputFile=None):
+    baseFileName = outputFile.rsplit("/")[-1].split(".")[0] if outputFile else None
```
PaulPrice: Use `os.path.basename`?

timj: Yes please. Something like: ` os.path.splitext(os.path.basename(outputFile))[0]` should do it.

PR #221, opened 2019-09-13
DM-17023: updates for changes to daf_butler and obs_base.

TallJimbo:


python/lsst/obs/subaru/gen3/hsc/instrument.py
```diff
@@ -66,27 +66,35 @@ def getName(cls):
...
+                "exposure_max": obsMax
+            }
+        )
+        registry.insertDimensionData(
+            "detector",
+            *[
+                {
+                    "instrument": self.getName(),
+                    "id": detector.getId(),
+                    "full_name": detector.getName(),
```
timj: 🎉 for having full name, raft name, and detector in raft name.

tests/test_ingest.py
```diff
@@ -26,7 +26,7 @@
 import os
 import lsst.utils.tests
 
-from lsst.obs.base.gen3.ingest_tests import IngestTestBase
+from lsst.obs.base.ingest_tests import IngestTestBase
 from lsst.obs.subaru.gen3.hsc import HyperSuprimeCam
 
 testDataPackage = "testdata_subaru"
```
timj: Not part of this change but it's been bugging me so can you please replace that `lsst.pex.exceptions.NotFoundError` with `LookupError` (it's an oddity of flake8 that it doesn't notice that we haven't really import pex_exceptions in this test and we are relying on others).

PR #226, opened 2019-10-23
DM-20574: Remove deprecated isr.doAddDistortionModel option.

czwa:


czwa:
Closing PR as `addDistortionModel` should be retained _through_ v19.

timj:
Note that in less than a month this change could be reapplied.

timj:
(also, why are these PRs showing that they are merged but also not in MERGED state but instead in CLOSED).

czwa:
My "look on the bright side" thought was that I have a very good idea of what needs to happen for the next deprecation in ip_isr+obs.
The closed vs merged thing is something github decided when I pushed the deletion of the commits.  I double checked that doAddDistortionModel was still there on master, and moved on.

PR #228, opened 2019-11-06
DM-21421 Add Configs for CharacterizeImageTask

natelust:
CharacterizeImageTask has a default name of characterizeImage, but
the config override file was named charImage reflecting the variable
name in ProcessCcdTask. This adds config files with the correct
names so that they can be appropriately discovered and loaded.

### DIFF
--- a/config/hsc/characterizeImage.py
+++ b/config/hsc/characterizeImage.py
@@ -1,0 +1,9 @@
+import os.path
+
+from lsst.utils import getPackageDir
+from lsst.meas.algorithms import ColorLimit
+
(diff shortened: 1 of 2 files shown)


config/hsc/characterizeImage.py
```diff
@@ -0,0 +1,9 @@
+import os.path
+
+from lsst.utils import getPackageDir
+from lsst.meas.algorithms import ColorLimit
+
+ObsConfigDir = os.path.join(getPackageDir("obs_subaru"), "config", "hsc")
+
+charImFile = os.path.join(ObsConfigDir, "charImage.py")
+config.load(charImFile)
```
andy-slac: Newline is missing from last line

PR #235, opened 2019-12-05
DM-21016: Handle DECam instrument signature data in gen3

parejkoj:


parejkoj:
This work was completed on DM-23976 and DM-23980 (though not via a cherry-pick of these commits).

PR #242, opened 2020-01-16
DM-23071: Reconfigure aperture correction source selector for bright stars

erykoff:


config/hsc/charImage.py
```diff
@@ -30,3 +30,14 @@
...
+config.measureApCorr.sourceSelector['science'].doFlags = True
+config.measureApCorr.sourceSelector['science'].doUnresolved = False
+config.measureApCorr.sourceSelector['science'].doSignalToNoise = True
+config.measureApCorr.sourceSelector['science'].flags.good = ['calib_psf_used']
+config.measureApCorr.sourceSelector['science'].flags.bad = []
+config.measureApCorr.sourceSelector['science'].signalToNoise.minimum = 200.0
+config.measureApCorr.sourceSelector['science'].signalToNoise.maximum = None
+config.measureApCorr.sourceSelector['science'].signalToNoise.fluxField = 'base_PsfFlux_instFlux'
+config.measureApCorr.sourceSelector['science'].signalToNoise.errField='base_PsfFlux_instFluxErr'
+config.measureApCorr.sourceSelector.name = 'science'
```
laurenam: Looks like you need spaces around the equals sign in 
`config.measureApCorr.sourceSelector['science'].signalToNoise.errField='base_PsfFlux_instFluxErr'`

It might also be nice to add a quick comment along the lines of:
# For aperture correction modeling, only use objects that were used in the PSF model and have S/N > 200

Also, VERY minor nitpick...most of (arghh...already internally inconsistent!) the rest of the file uses double quotes (and I was recently asked why pick one or the other...I tend to just always go double but try to be consistent with the rest of the file as is...not really possible here, so do as you wish!).

PS  This is not an official review, I was just curious what S/N threshold you landed on.  👍 on 200 :)

PR #245, opened 2020-01-17
DM-21308: Replace doApplyUberCal with doApplyExternal configs

erykoff:
Also update default configs.

### DIFF
--- a/python/lsst/obs/hsc/makeCoaddTempExp.py
+++ b/python/lsst/obs/hsc/makeCoaddTempExp.py
@@ -23,7 +23,6 @@
 
 from lsst.pipe.tasks.makeCoaddTempExp import MakeCoaddTempExpTask, MissingExposureError
 import lsst.daf.persistence as dafPersist
-import lsst.afw.image as afwImage
 import numpy as np
 
 __all__ = ["SubaruMakeCoaddTempExpTask"]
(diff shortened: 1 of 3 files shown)


python/lsst/obs/hsc/makeCoaddTempExp.py
```diff
@@ -53,10 +52,14 @@ def getCalibratedExposure(self, dataRef, bgSubtracted):
...
-        updated to the `jointcal_wcs`, otherwise it will be calibrated via the
-        Exposure's own PhotoCalib and have the original SkyWcs.
+
+        If config.doApplyExternalPhotoCalib is `True`, the photometric calibration
+        (`photoCalib`) is taken from `config.externalPhotoCalibName` via the
+        `name_photoCalib` dataset.  Otherwise, the photometric calibration is
+        retrieved from the `calexp`.  When `config.doApplyExternalSkyWcs` is `True`,
+        the astrometric calibration is taken from `config.externalSkyWcsName` with
+        the `name_wcs` dataset.  Otherwise, the astrometric calibration is taken
+        from the `calexp`.
```
mrawls: Please say "processed visit image" or "processed exposure" or something other than "calexp."

python/lsst/obs/hsc/makeCoaddTempExp.py
```diff
@@ -74,26 +77,15 @@ def getCalibratedExposure(self, dataRef, bgSubtracted):
...
-                skyWcs = dataRef.get("jointcal_wcs")
-                exposure.setWcs(skyWcs)
+        if self.config.doApplyExternalPhotoCalib:
+            photoCalib = dataRef.get(f"{self.config.externalPhotoCalibName}_photoCalib")
         else:
             photoCalib = exposure.getPhotoCalib()
 
+        if self.config.doApplyExternalSkyWcs:
+            skyWcs = dataRef.get(f"{self.config.externalSkyWcsName}_wcs")
+            exposure.setWcs(skyWcs)
```
mrawls: This is great - the exposure already has a WCS from the calexp (heh) if this doesn't happen, so it makes sense to only setWcs if we are overriding it.
Similarly, isn't it true that the exposure already has a photoCalib from the calexp? If so, then this can be simplified even further by removing the the else above (which retrieves the photoCalib from the exposure if there's no external one) and only calls setPhotoCalib if there is a new one.

jdswinbank:
I haven't looked carefully at what's going on here, so apologies if I've missed the point, but... 

While I'm seeing review comments changes to on `SubaruMakeCoaddTempExpTask`, I'm seeing that whole task being removed on https://github.com/lsst/obs_subaru/pull/224. At least at first glance, this seems ... unnecessary.

mrawls:
That's actually delightful! It turns out the comments here should apply instead to the parallel PR for pipe_tasks for DM-21308, so I will leave them for easy reference. (I looked at this PR before that one and got very confused about code duplication, so I'm thrilled it is going away)

erykoff:
@jdswinbank It is going away, yes.  And originally that other ticket (DM-20074) was a blocker for this one, but that has proved challenging so we decided to just get this one done in spite of the slight overhead of updating a mostly redundant function that is going to be removed soon.

PR #247, opened 2020-01-24
DM-22708: Fix decam gen3 ingest

parejkoj:


tests/test_ingest.py
```diff
@@ -42,15 +42,15 @@ def setUp(self):
         self.ingestDir = os.path.dirname(__file__)
         self.instrument = HyperSuprimeCam()
         self.file = os.path.join(testDataDirectory, "hsc", "raw", "HSCA90402512.fits.gz")
-        self.dataId = dict(instrument="HSC", exposure=904024, detector=50)
+        self.dataIds = [dict(instrument="HSC", exposure=904024, detector=50)]
```
timj: If you use `lsst.daf.butler.utils.iterable` in the testing base class then you won't need to change all these lines in the current tests and they can be straight dicts.

parejkoj: I'd already made the necessary changes to the other packages, and I think it's worth leaving so that it is very clear to users that they can test multiple dataIds.

PR #249, opened 2020-01-28
DM-20841: Support for approximating WCS as SIP

PaulPrice:


policy/HscMapper.yaml
```diff
@@ -121,7 +121,7 @@ exposures:
     tables:
     - raw
     - raw_visit
-    template: '%(pointing)05d/%(filter)s/corr/%(tract)04d/CALEXP-%(visit)07d-%(ccd)03d.fits'
+    template: '%(pointing)05d/%(filter)s/calibrated/%(tract)04d/CALEXP-%(visit)07d-%(ccd)03d.fits'
```
TallJimbo: While I realize these are more or less what Furusawa-san used as an example, I wonder if starting these paths with `%(tract)04d/calibrated` or `calibrated/%(tract)04d` or making them more like the `forced_src` paths (which do start with `%(pointing)05d` - unfortunately, I'd say) in how the tract is specified would be better.  Not a big deal either way - mostly I think it's hard to make these much better without changing more, and that's disruptive.

PR #254, opened 2020-02-11
DM-23352: Add jointcal, skyCorr to forcedPhotCcd

PaulPrice:


config/hsc/makeWarp.py
```diff
@@ -0,0 +1,11 @@
...
+from lsst.utils import getPackageDir
+
+# Load configs shared between assembleCoadd and makeCoaddTempExp
+config.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "coaddBase.py"))
+
+config.recalibrate.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "recalibrate.py"))
+
+config.doApplySkyCorr = True
+config.warpAndPsfMatch.warp.warpingKernelName = 'lanczos5'
+config.coaddPsf.warpingKernelName = 'lanczos5'
```
TallJimbo: Since `makeWarp` inherits from `makeCoaddTempExp`, maybe this should just load the config for that, to keep them in sync?

PR #263, opened 2020-03-27
DM-23976: Use the base class implementations for gen3 defects and camera

timj:


python/lsst/obs/subaru/instrument.py
```diff
@@ -49,13 +48,15 @@ class HyperSuprimeCam(Instrument):
...
+    policyName = "hsc"
+    obsDataPackage = "obs_subaru_data"
     filterDefinitions = HSC_FILTER_DEFINITIONS
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         packageDir = getPackageDir("obs_subaru")
         self.configPaths = [os.path.join(packageDir, "config"),
-                            os.path.join(packageDir, "config", "hsc")]
+                            os.path.join(packageDir, "config", self.policyName)]
```
parejkoj: Per my comment in the obs_base PR, `self.policyName` here could probably be replaced with `self.name.lower()`.

python/lsst/obs/subaru/instrument.py
```diff
@@ -216,48 +215,8 @@ def writeCuratedCalibrations(self, butler):
...
-                })
-        # Second loop actually does the inserts and filesystem writes.
-        with butler.transaction():
-            butler.registry.insertDimensionData("calibration_label", *dimensionRecords)
-            # TODO: vectorize these puts, once butler APIs for that become
-            # available.
-            for defect, dataId in datasetRecords:
-                butler.put(defect, datasetType, dataId)
+        # Write defects
+        self.writeStandardTextCuratedCalibrations(butler)
```
parejkoj: The comment is misleading, I think: it's writing more than just defects, it's whatever is in `standardCuratedDatasetTypes`.

python/lsst/obs/subaru/instrument.py
```diff
@@ -140,13 +141,11 @@ def writeCuratedCalibrations(self, butler):
         a standardized approach to this problem.
         """
 
-        # Write cameraGeom.Camera, with an infinite validity range.
-        datasetType = DatasetType("camera", ("instrument", "calibration_label"), "Camera",
-                                  universe=butler.registry.dimensions)
-        butler.registry.registerDatasetType(datasetType)
+        # Write cameraGeom.Camera
+        self.writeCameraGeom(butler)
```
parejkoj: I think the comment here is unnecessary, given the method name.

PR #264, opened 2020-04-02
DM-24062: Add tasks to generate parquet Source Table

yalsayyad:


config/calibrate.py
```diff
@@ -55,4 +55,5 @@
 config.deblend.maskLimits["NO_DATA"] = 0.25  # Ignore sources that are in the vignetted region
 config.deblend.maxFootprintArea = 10000
 
-config.measurement.plugins.names |= ["base_Jacobian", "base_FPPosition"]
+config.measurement.plugins.names |= ["base_Jacobian", "base_FPPosition",
```
morriscb, line 58: What do you think? Put this in the config or the default on the task? Right now the `base_Local`s are defaults in the tasks multiBand and imageDiference.

yalsayyad: The thread here hasn't really helped decide one way or another: https://lsstc.slack.com/archives/C2JPMCF5X/p1585871128255400  People took to questioning Yes or No instead of A or B.  

Currently, obs_subaru is the only place a `Source.yaml` exists which is the immediate user of these columns.  I will either (a) keep it in obs_subaru and add a "TO DO: remove on DM-XXXXX" where DM-XXXXX is the ticket to write a obs_lsst/imsim/Source.yaml. while  HSC folks are the guinea pigs in the meantime or (b) move it to `calibrateConfig.setDefaults` now. Concern there is that when people start using the columns, they'll be impossible to remove later if we start calibrating catalogs differently.

yalsayyad: @morriscb What's your preference?

morriscb: I mean, you say HSC is the only user of this currently but part of my work in creating these was to get the SDM running in AP which is currently DECam data. Having these values that are kind of fundamental to producing calibrated data products with the SDM be per obs package seems like a bunch of duplicated code that we don't need. If you want to make this a new ticket that's fine but I would need them to be done sooner rather than later.

yalsayyad: I'll move it to calibrate now on this ticket since you need it for DECam AP!  Thanks for letting me know.

PR #267, opened 2020-04-10
DM-15536: adapt to moving visit definition out of ingest

TallJimbo:


config/hsc/convertRepo.py
```diff
@@ -18,7 +19,8 @@
 # skymap.
 config.rootSkyMapName = "hsc_rings_v1"
 
-config.raws.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "ingest-gen3.py"))
+config.defineVisits.load(os.path.join(getPackageDir("obs_subaru"), "config", "hsc", "defineVisits.py"))
```
timj, line 22: Use `__file__` instead of `getPackageDir`.

config/hsc/convertRepo.py
```diff
@@ -6,6 +6,7 @@
 # probably won't survive persisting the config.
 import lsst.obs.subaru.gen2to3
 
+config.instrument = "lsst.obs.subaru.HyperSuprimeCam"
```
timj, line 9: Why is this set in this config but the DECam equivalent doesn't set instrument?

TallJimbo: I hadn't noticed that DECam didn't because the command-line script code sets it.  I think I should probably set it in DECam in order to make the `Task` more usable directly (though that's true of a lot of the logic in the script, I think).

In any case, it's an example of the whole mess where you have to specify an instrument to get the obs-package overrides that then set the instrument.  I think that's one of the problems that would go away if we move `writeCuratedCalibrations` and instrument registration out of `ConvertRepoTask`, because those are the only reasons we need the `Instrument` there.

config/hsc/convertRepo.py
```diff
@@ -15,7 +16,8 @@
 # skymap.
 config.rootSkyMapName = "hsc_rings_v1"
 
-config.raws.load(os.path.join(os.path.dirname(__file__), "ingest-gen3.py"))
+config.defineVisits.load(os.path.join(os.path.join(os.path.dirname(__file__), "defineVisits.py"))
```
timj, line 19: Missing a closing bracket I think.

PR #270, opened 2020-04-15
DM-24466: Can't run processCcd on multiple CCDs

plazas:


python/lsst/obs/hsc/hscMapper.py
```diff
@@ -230,11 +226,11 @@ def map_linearizer(self, dataId, write=False):
             storage=self.rootStorage)
 
     def bypass_linearizer(self, datasetType, pythonType, butlerLocation, dataId):
-        """Return the linearizer.
+        """Return a Linearizer() for a particular detector. Linearizer() applies to
+        one detector only (i.e., it is detector-specific). When processing multiple
+        detectors, a new instance of Lineaizer() will be created each time.
```
jdswinbank, line 231: Could you please make this docstring follow the [numpydoc guidelines](https://developer.lsst.io/python/numpydoc.html#py-docstring-sections)?

It would also be helpful to include a little more information about the `Linearizer` that is being returned.

Specifically, I'd suggest something like:

```python
"""Return a linearizer for the given detector.

On each call, a fresh instance of `Linearizer` is returned; the caller is responsible for initializing it
appropriately for the detector.

Parameters
----------
`datasetType`: ...
`pythonType`: ...
`butlerLocation`: ...
`dataId`: ...

Notes
-----
Linearizers are not saved to persistent storage; rather, they are managed entirely in memory.
On each call, this function will return a new instance of `Linearizer`, which must be managed
(including setting it up for use with a particular detector) by the caller. Calling
`bypass_linearizer` twice for the same detector will return _different_ instances of `Linearizer`,
which share no state.
"""
```

tests/test_repository.py
```diff
@@ -183,7 +183,11 @@ def testFlat(self):
         self.assertEqual(flat.getDetector().getId(), self.ccdNum)
 
     def testLinearizer(self):
-        self.assertTrue(self.butler.get('linearizer', ccdnum=1))
+        lin1 = self.butler.get('linearizer', ccdnum=1)
+        lin2 = self.butler.get('linearizer', ccdnum=2)
+        self.assertTrue(lin1)
+        self.assertTrue(lin2)
```
jdswinbank, line 189: I'm not sure these `assertTrue`s mean much — what does it mean for a `Linearizer` to be `True`?

If you just want to check that you have an object of some sort, I'd suggest `assertIsNotNone`.

timj: Aren't you wanting to test that the instances are distinct? (so checking that `id(lin1) != id(line2)` ?

jdswinbank: that's the next line...

plazas: I can change it to `assertIsNotNone`. The original code in this test (before my modifications) had `assertTrue`.

PR #275, opened 2020-04-30
Add relevant datatypes and configs for processCcdWithFakesDriver

sr525:
Add datatypes for `fakes_source` and
    processCcdWithFakesDriver's metadata as well as a config file
    for processCcdWithFakesDriver.

### DIFF
--- a/config/hsc/processCcdWithFakesDriver.py
+++ b/config/hsc/processCcdWithFakesDriver.py
@@ -1,0 +1,10 @@
+"""
+hsc-specific overrides for processCcdWithFakesDriverTask
+"""
+import os.path
+
+
+config.doMakeSourceTable = True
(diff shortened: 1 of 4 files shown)


config/hsc/processCcdWithFakesDriver.py
```diff
@@ -0,0 +1,10 @@
+"""
+hsc-specific overrides for processCcdWithFakesDriverTask
+"""
+import os.path
+
+
+config.doMakeSourceTable = True
+#config.processCcdWithFakes.load(os.path.join(os.path.dirname(__file__), "processCcdWithFakes.py"))
```
czwa, line 8: Remove commented lines.

sr525: Good catch, this actually was waiting on another ticket being merged so that it could work. Looking into this made me realise that I was missing a processCcdWithFakes.py in obs_subaru/config/hsc/.

PR #277, opened 2020-05-01
DM-24664: Refactor data ID translation rule system

TallJimbo:


python/lsst/obs/subaru/instrument.py
```diff
@@ -45,6 +46,22 @@
 log = logging.getLogger(__name__)
 
 
+class _HscAbstractFilterKeyHandler(KeyHandler):
+    """KeyHandler for HSC filter keys that should be mapped to AbstractFilters.
```
parejkoj, line 50: Is this materially different from `PhysicalToAbstractFilterKeyHandler`? I don't remember why I kept it around when I was doing the original converter cleanup (if there was a reason beyond that I just missed it), but I think it can just be replaced by that.

TallJimbo: Yup, looks like they're identical.  Will drop this one.

PR #281, opened 2020-05-14
DM-24946: Propagate detectCoaddSources configs to multiBandDriver

leeskelvin:


config/multiBandDriver.py
```diff
@@ -3,7 +3,7 @@
 import os.path
 
 
-for sub in ("mergeCoaddDetections", "measureCoaddSources", "mergeCoaddMeasurements", "forcedPhotCoadd"):
+for sub in ("mergeCoaddDetections", "measureCoaddSources", "mergeCoaddMeasurements", "forcedPhotCoadd", "detectCoaddSources"):
```
laurenam, line 6: I may have put it at the front of the list to reflect the order of operation in multiband.

config/hsc/multiBandDriver.py
```diff
@@ -2,7 +2,7 @@
 import os.path
 
 
-for sub in ("mergeCoaddDetections", "measureCoaddSources", "mergeCoaddMeasurements", "forcedPhotCoadd"):
+for sub in ("mergeCoaddDetections", "measureCoaddSources", "mergeCoaddMeasurements", "forcedPhotCoadd", "detectCoaddSources"):
```
laurenam, line 5: I don’t see a `detectCoaddSources.py` in `config/hsc`, but I suppose it doesn’t hurt to add it here given the if clause.

PR #282, opened 2020-05-15
DM-24923: eliminate redundant class name & name requirement in butler command test cases

n8pease:


tests/test_ingest.py
```diff
@@ -42,25 +42,29 @@ class HscIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
-        self.dataIds = [dict(instrument="HSC", exposure=904024, detector=50)]
-
-        super().setUp()
-        self.butler = Butler(self.root, run=self.outputRun)
-        self.visits = {
+    ingestDir = os.path.dirname(__file__)
+
+    instrumentClassName = "lsst.obs.subaru.HyperSuprimeCam"
+
+    instrumentName = "HSC"
```
timj, line 49: I think this can go now can't it?

tests/test_ingest.py
```diff
@@ -42,25 +42,29 @@ class HscIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
-        self.instrument = "lsst.obs.subaru.HyperSuprimeCam"
-        self.instrumentName = "HSC"
-        self.file = os.path.join(testDataDirectory, "hsc", "raw", "HSCA90402512.fits.gz")
-        self.dataIds = [dict(instrument="HSC", exposure=904024, detector=50)]
-
-        super().setUp()
-        self.butler = Butler(self.root, run=self.outputRun)
-        self.visits = {
+    ingestDir = os.path.dirname(__file__)
+
```
timj, line 46: I'm not sure I'm in favor of all the blank lines between properties.

tests/test_ingest.py
```diff
@@ -42,25 +42,29 @@ class HscIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
+    instrumentClassName = "lsst.obs.subaru.HyperSuprimeCam"
+
+    instrumentName = "HSC"
+
+    file = os.path.join(testDataDirectory, "hsc", "raw", "HSCA90402512.fits.gz")
```
timj, line 51: This .gz reminds me that I think we want to change the default for ingest to also include `.fits.gz` and `.fits.fz`.

tests/test_ingest.py
```diff
@@ -69,7 +73,8 @@ def checkRepo(self, files=None):
         # We ignore `files` because there's only one raw file in
         # testdata_subaru, and we know it's a science frame.
         # If we ever add more, this test will need to change.
-        expanded = self.butler.registry.expandDataId(self.dataIds[0])
+        butler = Butler(self.root, run=self.outputRun)
```
timj, line 76: I think this should be a readonly butler as above.

tests/test_ingest.py
```diff
@@ -42,25 +42,29 @@ class HscIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
...
+
+    instrumentName = "HSC"
+
+    file = os.path.join(testDataDirectory, "hsc", "raw", "HSCA90402512.fits.gz")
+
+    dataIds = [dict(instrument="HSC", exposure=904024, detector=50)]
+
+    @property
+    def visits(self):
+        butler = Butler(self.root, run=self.outputRun)
```
timj, line 57: I think this should be:
```python
butler = Butler(self.root, collections=[self.outputRun])
```
because you want it to be read only and you aren't writing any data into it.

tests/test_ingest.py
n8pease: @timj did you want me to do something about this in this ticket? (FWIW if it involves more than a line or 2 of code I'd rather ticket it)

timj: It was one line in python/lsst/obs/base/script/ingestRaws.py where we check file extension and could include the .gz/.fz there directly -- the ingestion clearly supports .fits.gz because the above file is a .fits.gz.  What I'm now wondering is why we are specifying a file here at all when it seems like we could specify the directory -- maybe this is deliberate to ensure that we know exactly which dataIds we will end up with in the butler.

TallJimbo: We should take files as any number of extra positional arguments and let the shell do the globbing for us.

PR #283, opened 2020-05-18
DM-23983: Cannot apply crosstalk in Gen 3 DECam processing

czwa:
Add dummy crosstalk mapper.

### DIFF
--- a/python/lsst/obs/hsc/hscMapper.py
+++ b/python/lsst/obs/hsc/hscMapper.py
@@ -226,7 +226,12 @@ class HscMapper(CameraMapper):
             storage=self.rootStorage)
 
     def map_crosstalk(self, dataId, write=False):
-        """Construct this on-the-fly in Gen2.
+        """Fake the mapping for crosstalk.
+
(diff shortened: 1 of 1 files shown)


python/lsst/obs/hsc/hscMapper.py
```diff
@@ -225,6 +225,11 @@ def map_linearizer(self, dataId, write=False):
             mapper=self,
             storage=self.rootStorage)
 
+    def map_crosstalk(self, dataId, write=False):
+        """Construct this on-the-fly in Gen2.
```
mrawls, line 229: Construct ... what on the fly? A bit more explanation here, please.

czwa: Updated docstring to be clearer:

Fake the mapping for crosstalk.
Crosstalk is constructed from config parameters, but we need Gen2 butlers to be able to respond to requests for it.  Returning None provides a response that can be used with the config parameters to generate the appropriate calibration.

PR #296, opened 2020-06-15
DM-24319: Add support for new parquet source tables in fgcmcal

erykoff:


config/hsc/fgcmBuildStarsTable.py
```diff
@@ -0,0 +1,29 @@
...
+config.densityCutMaxPerPixel = 2000
+config.filterMap = {'g': 'g', 'r': 'r', 'r2': 'r', 'i': 'i', 'i2': 'i',
+                    'z': 'z', 'y': 'y',
+                    'N387': 'N387', 'N816': 'N816', 'N921': 'N921',
+                    'N1010': 'N1010'}
+config.primaryBands = ('i', 'r', 'g', 'z', 'y', 'N387', 'N816', 'N921', 'N1010')
+config.doSubtractLocalBackground = True
+
+config.fgcmLoadReferenceCatalog.refObjLoader.ref_dataset_name = 'ps1_pv3_3pi_20170110'
+config.fgcmLoadReferenceCatalog.refFilterMap = {'g': 'g', 'r': 'r', 'r2': 'r',
```
parejkoj, line 18: I was confused by this line and the above `filterMap` because I believe our LoadReferenceObjectsTask does not need matching filter names to be defined in the `filterMap` (e.g. how our PS1 refcats work, because most filter names exist in both). Maybe add a comment to that effect, because this is not the same thing?

config/hsc/fgcmCalibrateTractTable.py
```diff
@@ -0,0 +1,18 @@
+import os.path
+
+config.fgcmBuildStars.load(os.path.join(os.path.dirname(__file__),
+                                        'fgcmBuildStarsTable.py'))
+config.fgcmFitCycle.load(os.path.join(os.path.dirname(__file__),
+                                      'fgcmFitCycle.py'))
+config.fgcmOutputProducts.load(os.path.join(os.path.dirname(__file__),
+                                            'fgcmOutputProducts.py'))
+config.fgcmFitCycle.aperCorrFitNBins = 0
+config.fgcmFitCycle.useRepeatabilityForExpGrayCutsDict = {'N387': True,
```
parejkoj, line 10: Is this "all filters True" or "just these specific filters True"? It's not clear from how the list is made: maybe there's a better way to make that clear? I don't know if using the `FilterDefinition`s in `hscFilters.py` might help?

PR #300, opened 2020-07-02
DM-25689: fixes for Gen2->Gen3 repo conversion

TallJimbo:


python/lsst/obs/subaru/strayLight/formatter.py
```diff
@@ -29,6 +29,7 @@
 class SubaruStrayLightDataFormatter(FileFormatter):
     """Gen3 Butler Formatters for HSC y-band stray light correction data.
     """
+    extension = ".fits"
```
laurenam, line 32: Can’t help but be curious...is the stutter (with L34) deliberate?

timj: No. I fixed it last week and the rebase didn’t notice so added the second one.

laurenam: Doh...!

TallJimbo: Gah, I saw this before merge, amended the commit to drop that part, and then somehow it came back before I merged.

PR #306, opened 2020-08-13
DM-26317: Add lru_cache for gen3 default camera

timj:


python/lsst/obs/subaru/_instrument.py
```diff
@@ -116,6 +118,11 @@ def getCamera(self):
         a standardized approach to writing versioned cameras to a Gen3 repo.
         """
         path = os.path.join(getPackageDir("obs_subaru"), self.policyName, "camera")
+        return self._getCameraFromPath(path)
+
+    @classmethod
```
SimonKrughoff, line 123: Same question as on the decam PR.

PR #309, opened 2020-08-21
DM-25156: Declare curated calibration types in instrument class, not config

timj:


python/lsst/obs/subaru/_instrument.py
```diff
@@ -55,6 +55,8 @@ class HyperSuprimeCam(Instrument):
     policyName = "hsc"
     obsDataPackage = "obs_subaru_data"
     filterDefinitions = HSC_FILTER_DEFINITIONS
+    additionalCuratedDatasetTypes = ("bfKernel", "transmission_optics", "transmission_sensor",
+                                     "transmission_filter", "transmission_atmosphere", "yBackground")
```
kfindeisen, line 59: It seems a bit odd that we've all been doing HSC work without needing this info before. Doesn't `ci_hsc` use these datasets? Is there a config somewhere else that needs to be updated?

timj: These were all listed explicitly in the obs_base default config for conversion...

timj:
What else did you expect to find in obs_subaru?

python/lsst/obs/subaru/_instrument.py
RobertLuptonTheGood: HSC doesn't separate these terms (although it'll be nice to have them for LSSTCam).  One point: I'm not sure that `transmission_atmosphere` belongs here as it's per-visit.  Also, it's a little odd to have `transmission_sensor` which is probably meant to be `1 - QE`, but it's a little unclear in the IR (around 950nm and redder) as the CCD becomes transparent.
Who defined this set of parameters?

timj: These are the calibrations that the HyperSuprimeCam gen3 instrument writes to the calibration collection when it is asked to write curated calibrations.

RobertLuptonTheGood: Someone must have chosen them, though.  And they don't make a lot of sense;  I'd be sorry to see them get baked into gen3 or HSC.

TallJimbo: My thinking was to leave this as-is until Gen2 is gone, not use Gen3 to intentionally break it.

PR #317, opened 2020-09-26
DM-25849: Update HSC to use jointcal refcat defaults

parejkoj:


yalsayyad:
Look's good.  Will you change the commit message to something a little more dramatic, so that it's obvious this is a big behavior change for our DRPs. Something like:

"Use default refcat in jointcal: Gaia DR2"

I'd prob just scroll past "Clean up jointcal settings" thinking "prob not important"

yalsayyad (approved):

PR #318, opened 2020-10-01
DM-26862: Add default config for new fgcm focal plane fitting

erykoff:


config/hsc/fgcmFitCycle.py
```diff
@@ -73,6 +73,17 @@
...
+                                'g': True,
+                                'r': True,
+                                'i': True,
+                                'N816': True,
+                                'z': True,
+                                'N921': True,
+                                'y': True,
+                                'N1010': True}
+config.ccdGrayFocalPlaneFitMinCcd = 50
+config.ccdGrayFocalPlaneChebyshevOrder = 3
```
laurenam, line 86: 3 is the default, no?

config/hsc/fgcmFitCycle.py
```diff
@@ -73,6 +73,17 @@
...
                             'N1010': False}
+config.ccdGrayFocalPlaneDict = {'N387': True,
+                                'g': True,
+                                'r': True,
+                                'i': True,
+                                'N816': True,
+                                'z': True,
+                                'N921': True,
+                                'y': True,
+                                'N1010': True}
```
laurenam, line 84: At first, I was going to say that if all these filter Dicts are meant to stay in sync, it might be useful to define one and set each to that (will make updating easier/less error prone going forward).  However, I do see that `N387` and `N1010` are set to `True` here and `False` ...so I guess I'm now just making sure that was intentional?!

config/hsc/fgcmFitCycle.py
```diff
@@ -73,6 +73,17 @@
...
+config.ccdGrayFocalPlaneFitMinCcd = 50
```
laurenam, line 85: Some comment on this choice would be helpful...where do the tradeoffs lie with slightly fewer CCDs?

laurenam (approved):

config/hsc/fgcmFitCycle.py
erykoff: These are intentional based on trial and error, that the shallow narrow-band filters just don't have enough data to fit reliably sub-ccd on a _per_ ccd level.  However, for the full focal plane fitting, these can be fit because the spatial scales are different. I'll add comments to the config.

PR #322, opened 2020-10-15
DM-24961: Remove SuprimeCam

timj:


PaulPrice (approved):

timj:
@PaulPrice -- I just found the HscSimMapper thing as well so I'm going to delete that whilst I'm here.

I'll also push a docstring cleanup to make the code compliant with current style.

PR #327, opened 2020-11-10
DM-27168: Phase out use of FilterProperty

czwa:
End usage of FilterProperty, as per RFC-730.


tests/test_repository.py
```diff
@@ -120,7 +120,7 @@ def testRaw(self):
         raw = self.butler.get("raw", visit=self.visit, ccd=self.ccdNum)
         ccd = raw.getDetector()
         self.assertEqual(raw.getDimensions(), geom.Extent2I(*self.rawSize))
-        self.assertEqual(raw.getFilter().getFilterProperty().getName(), self.filter)
+        self.assertEqual(raw.getFilter().getName(), self.filter)
```
kfindeisen, line 123: As in lsst/pipe_tasks#428, I'm pretty sure this is an idiom for `getCanonicalName()` that predates its introduction.

kfindeisen (approved):

PR #333, opened 2021-01-02
DM-27170: Deprecate and phase out afw::image::Filter

parejkoj:


bin.src/showDither.py
```diff
@@ -40,7 +40,7 @@ def main(butler, visits, fields, fieldRadius, showCCDs=False, aitoff=False, alph
         visits.append(v)
         ra.append(sky[0].asDegrees())
         dec.append(sky[1].asDegrees())
-        filters[v] = exp.getFilter().getName()
+        filters[v] = exp.getFilterLabel().bandLabel
 
     plt.clf()
     if aitoff:
```
kfindeisen, line 46: This looks fine by itself, but in the code below there's a dictionary that expects `r1` as one of the filters.

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -41,16 +41,14 @@ def getLongFilterName(short):
...
         return short
     if short.startswith("NB") or short.startswith("IB"):
         num = int(short[2:].lstrip("0"))
         return "%s%04d" % (short[:2], num)
-    f = Filter(short)
-    for a in f.getAliases():
-        if a.startswith("HSC") or a.startswith("NB") or a.startswith("IB"):
-            return a
+    for filter in hscFilters.HSC_FILTER_DEFINITIONS:
+        if short == filter.afw_name or short == filter.band:
```
kfindeisen, line 50: I thought we were planning to deprecate `afw_name`. Is that not possible, or is this a temporary fix?

tests/test_ingest.py
```diff
@@ -44,6 +45,7 @@ class HscIngestTestCase(IngestTestBase, lsst.utils.tests.TestCase):
                                       "transmission_sensor", "transmission_filter", "transmission_atmosphere")
     ingestDir = os.path.dirname(__file__)
     instrumentClassName = "lsst.obs.subaru.HyperSuprimeCam"
+    filterLabel = lsst.afw.image.FilterLabel(physical="HSC-I", band="i")
```
kfindeisen, line 48: This appears to be unused.

kfindeisen (changes requested):
I don't think these changes are complete, as `obs_subaru` seems to have a lot of hard-coded filter names that don't appear on a search for `getFilter`. In particular, https://github.com/lsst/obs_subaru/blob/tickets/DM-27170/python/lsst/obs/subaru/filterFraction.py still relies on `CoaddInputs` using the `afw_name`, which will not be the case once lsst/pipe_tasks#448 merges. I think you need to use the physical name in at least some places to ensure HSC-R and HSC-R2 are distinguishable.

tests/test_ingest.py
kfindeisen: Never mind, forgot this is defined by `IngestTestBase`.

bin.src/showDither.py
parejkoj: Ugh. I was hesitant to even touch this file, since it's an executable from 2013 that probably doesn't work anyway. I'm not sure what to do here: there is no good solution and I cannot test this code. Do we know if it even works?

parejkoj: Actually, that's fine: `ENG-R1` has band `r1`, so `bandLabel` is absolutely correct here.

tests/test_ingest.py
parejkoj: The parent class `IngestTestBase` uses it.

python/lsst/obs/hsc/makeTransmissionCurves.py
parejkoj: My change here was to get rid of the call to `Filter` and instead just use the FilterDefinitions. I've filed DM-28334 about this issue: the correct solution is to relabel the transmission curves with the physical filter name.

parejkoj:
I've pushed what I believe is a fix to the `FilterFraction` question you raised above.

kfindeisen (approved):
Looks good, assuming there aren't any more hardcoded strings lurking around. Unfortunately, I can't check that from GitHub.

bin.src/showDither.py
kfindeisen: Yes: https://github.com/lsst/obs_subaru/blob/master/python/lsst/obs/hsc/hscFilters.py#L63-L64

yalsayyad (approved):

PR #335, opened 2021-01-22
DM-28389: Create a gen3 task to collate tract-level parquet tables for QA analyses

laurenam:


pipelines/makeQaTractTables.yaml
```diff
@@ -0,0 +1,6 @@
+description: Make collated tract-level QA tables for use in plotting pipelines
+tasks:
+  makeQaTractTables:
+    class: lsst.analysis.drp.makeQaTractTables.MakeQaTractTablesTask
+    config:
+       filterMap: {"g": "HSC-G", "r": "HSC-R", "i": "HSC-I", "z": "HSC-Z", "y": "HSC-Y", "N921": "NB0921"}
```
timj, line 6: It seems wrong to me that we have to duplicate the mapping from band to physical filter here (and it doesn't have the second physical filters) since this information is already defined explicitly in obs_subaru in hscFilters.py and also directly discoverable from the butler itself.

laurenam: Oh, that’s great!  I would be very happy to be getting this from the butler and get rid of this file altogether 😄

timj: You can use `butler.registry.queryDimensionRecords`. From the command line you get something like:

```
$ butler query-dimension-records tmp physical_filter
instrument              name                          band            
---------- ----------------------------- -----------------------------
       HSC                        ENG-R1                            r1
       HSC                         HSC-G                             g
       HSC                         HSC-I                             i
       HSC                        HSC-I2                             i
       HSC                         HSC-R                             r
       HSC                        HSC-R2                             r
       HSC                         HSC-Y                             y
       HSC                         HSC-Z                             z
       HSC                        IB0945                          I945
       HSC                        NB0387                          N387
       HSC                        NB0400                          N400
       HSC                        NB0468                          N468
       HSC                        NB0515                          N515
       HSC                        NB0527                          N527
       HSC                        NB0656                          N656
       HSC                        NB0718                          N718
       HSC                        NB0816                          N816
       HSC                        NB0921                          N921
       HSC                        NB0926                          N926
       HSC                        NB0973                          N973
       HSC                        NB1010                         N1010
       HSC                          NONE                  UNRECOGNISED
       HSC                            PH                            PH
       HSC                            SH                            SH
```

laurenam: Perfect...thanks!

timj: You can restrict the query to just the filter you are interested in for the instrument you use.

TallJimbo: We don't want to run `queryDimensionRecords` from within a running `PipelineTask`, though, and I'm not sure you'd even be able to.  The information you want will probably be in the data IDs the `PipelineTask` is given; when I dive into the code I'll see if I can figure out exactly what you want and provide an example of how to get it.

timj: Okay. I didn't know the context. The DataId should also have band in it of course.

laurenam: Ok, it seems I should be more explicit about the context and issue here.  The problem is that I am reading in the `deepCoadd_obj` tables (persisted by the `postprocess` task in `pipe_tasks`).  These are multilevel DataFrames that have a "filter" level...this level is (currently) indexed on the HSC *physical* filter names, e.g.
```
In [152]: from lsst.daf.butler import Butler
In [153]: rootDir = "/project/hsc/gen3repo/rc2w02_ssw03"
     ...: butler = Butler(rootDir, collections="HSC/runs/RC2/w_2021_02")
In [154]: obj = butler.get("deepCoadd_obj", tract=9813, patch=40)
In [155]: obj
Out[155]: 
dataset                                    forced_src  ...     ref
filter                                          HSC-G  ...  NB0921
column            base_CircularApertureFlux_12_0_flag  ... patchId
id                                                     ...        
43158597349146625                                True  ...     4,4
43158597349146626                                True  ...     4,4
...                                               ...  ...     ...
43158597349192297                               False  ...     4,4
43158597349192298                               False  ...     4,4
[32117 rows x 13674 columns]

In [156]: obj.columns.get_level_values("filter").unique()
Out[156]: Index(['HSC-G', 'HSC-R', 'HSC-I', 'HSC-Z', 'HSC-Y', 'NB0921'], dtype='object', name='filter')
```
I need to access the filter levels associated with the "band" that is in the dataId supplied to my `pipetask` command line.  Now, it turns out I can access Tim's suggested lookup from within the tasks `runQuantum()` method as follows:
```
band = butlerQC.quantum.outputs[outputCatalogs][0].dataId["band"]
physicalFilterGtor = butlerQC.registry.queryDimensionRecords("physical_filter", dataId={"band": band})
for physicalFilter in physicalFilterGtor:
    print(physicalFilter)
    print(physcalFilter.name)
```
this gives:
```
physical_filter:
  instrument: 'HSC'
  name: 'HSC-R'
  band: 'r'
HSC-R
physical_filter:
  instrument: 'HSC'
  name: 'HSC-R2'
  band: 'r'
HSC-R2
```
That is actually highlights a problem (perhaps one reason behind Jim's advice to NOT use this from within a `PipelineTask`!)  as there can be multiple physical filters for a given camera associated with a single generic band.
So...I *think* I am still stuck with my `filterMap` config here (and I do imagine it will become unnecessary once the `postprocess` scripts get updated to Gen3 and the index will then be on band?)

Please let me know if I am still missing something, or if there is a better way to get this (temporary?) mapping.

timj: I'm a bit confused because by definition there can be multiple physical filters corresponding to a single band and surely this has to be handled?

TallJimbo: Ok, I hadn't realize this problem was present in the `deepCoadd_obj` schemas (even when commenting on the `analysis_drp` PR a moment ago).  We need to get that fixed; it may not have been considered a problem back when we first defined them, because then we had little or no consistency in which filter labels we used, but it's definitely a bug now.

I _suspect_ this will just be fixed automatically when we create `deepCoadd_obj` in Gen3, but since right now we're mostly working with the Gen2->Gen3 converted ones, we need to fix that one to.  There's some chance that will have been recently done by the `FilterLabel` work, but I don't want to count on it.

natelust: I randomly saw this email in my inbox, I apologize about adding another cook to the kitchen, but I have a few thoughts on this.

I regret making the registry a public attribute on ButlerQC, and I think we should change it to private. As such I do not think it should be used here. There are two things that come to mind. First, though unlikely it is possible that someone changes something in the registry during an active run, which would mean different executions of `runQuantum` would get different answers from the registry.

Secondly, There would be no way to predict ahead of time which info one might need from the registry, and thus our scheme of "know everything" before a run and have a stripped down registry to use for a run would go out of the window. We should stop allowing dynamic access at all during a run.

The behavior above could be emulated with some changes to connections classes, but I don't think it should be. The configuration a pipeline is run with should be a fixed quantity that lives with the pipeline (or in a configuration file) it should not be subject to dynamic look-ups. It is possible that someone will want to configure a different set of mapping that what would be reported by the registry. 

As far as duplication within a single pipeline file, there is a mechanism to get around this see the following:
```

description: Make collated tract-level QA tables for use in plotting pipelines
parameters:
  filterMap: {"g": "HSC-G", "r": "HSC-R", "i": "HSC-I", "z": "HSC-Z", "y": "HSC-Y", "N921": "NB0921"}
tasks:
  makeForcedQaTractTables:
     class: lsst.analysis.drp.makeQaTractTables.MakeForcedQaTractTablesTask
     config:
        filterMap: parameters.filterMap
  makeUnforcedQaTractTables:
     class: lsst.analysis.drp.makeQaTractTables.MakeUnforcedQaTractTablesTask
     config:
       filterMap: parameters.fitlerMap
```

TallJimbo: Just created [DM-28479](https://jira.lsstcorp.org/browse/DM-28475).  I'll do a bit of investigation there before trying to figure out who to hand it off ot.

laurenam: Exactly... I think the confusion was the thought that I was trying to switch to `physical` rather than sticking with `band`.  This is not the case at all (I’m totally on board with just `band` for any and all things coadd!). It’s just that I currently need `physical` to get at the data in the `_ojb` tables.

I have also been speculating that this’ll just naturally get fixed along with the Gen3-ification of the code making the `_obj` tables, hence me very much referring to this as a temporary fix.  I’m perfectly happy to hold off merging this until the fix is in so that this never shows up as part of the history on master (it shouldn’t be too painful for me to just run off the two branches manually for the intervening RC2 runs).

laurenam: Thanks, Nate. Yeah, I fully landed on NOT using `ButlerQC.regitry`, but it did seem worthwhile pointing out that it could be done (and hence a trap others could fall into).

Ah, yes, I knew there had to be a way to reuse that config...thanks for the tip!

timj: It still seems fundamentally wrong to me that a pipeline file should define a filter-to-band mapping when we already have that mapping defined in two other places (such as in obs_subaru itself in the filter definitions).

natelust: It depends on what it is. I see nothing wrong with having a config file that can be imported and used anywhere you want "here is what is defined in obs_subaru" But also these are configurable fields, so if someone wants to configure them in a non-standard way for whatever their reason (emulate a wider band from a different camera etc) then they should be able to specify the config parameter to their liking.

TallJimbo: > It still seems fundamentally wrong to me that a pipeline file should define a filter-to-band mapping when we already have that mapping defined in two other places (such as in obs_subaru itself in the filter definitions).

It's not a `physical_filter`-to-`band` mapping; it's a `band`-to-`pseudo/historical-physical_filter` mapping, and we don't have one of those in obs_subaru because it's an ill-defined thing.  Unfortunately it's exactly the ill-defined thing that's still in use in Gen2.

PR #337, opened 2021-01-29
DM-27173: Add doc field to obs.base.FilterDefinition

parejkoj:


python/lsst/obs/hsc/hscFilters.py
```diff
@@ -39,12 +39,16 @@
                      lambdaEff=477, alias={'W-S-G+'}),
     FilterDefinition(physical_filter="HSC-R",
                      band="r",
+                     doc="Original r-band filter, replaced in July 2016 with HSC-R2.",
                      lambdaEff=623, alias={'W-S-R+'}),
     FilterDefinition(physical_filter="ENG-R1",
                      band="r1",
+                     doc="A filter used during early camera construction;"
+                         "essentially no data was taken with this filter.",
```
timj, line 47: I'm not really a fan of "essentially no data". How about "very little data" ?

timj (approved):

python/lsst/obs/hsc/hscFilters.py
parejkoj: I'm just quoting from RHL in DM-28460.

PR #344, opened 2021-02-23
DM-28759: Add config for SourceTableDriver

yalsayyad:


timj:
My understanding is that this is a gen2 PR and we can close it.

PR #352, opened 2021-03-10
DM-28597: Fix colorterm/photoCal filterLabel confusion

parejkoj:


config/colorterms.py
```diff
@@ -1,52 +1,14 @@
...
-        'i': Colorterm(primary="i", secondary="i"),
-        'z': Colorterm(primary="z", secondary="z"),
-        'y': Colorterm(primary="y", secondary="y"),
-        # Names used by data IDs in both Gen2 and Gen3, and
-        # Exposure.getFilter() in Gen3 (data is the same).
-        'HSC-G': Colorterm(primary="g", secondary="g"),
-        'HSC-R': Colorterm(primary="r", secondary="r"),
-        'HSC-I': Colorterm(primary="i", secondary="i"),
-        'HSC-Z': Colorterm(primary="z", secondary="z"),
-        'HSC-Y': Colorterm(primary="y", secondary="y"),
```
erykoff, line 17: I don't agree with removing the HSC -> HSC no-op option.  I think that for convenience it's nice to be able to calibrate an HSC image against an HSC reference catalog, and let it do the straight pass-through when you leave color terms "on".
Though we then have to add on all the physical filters (including NB) to this list.

erykoff (commented):
I would like to retain the no-op hsc/hsc option, though we need to add all the physical filters.

config/colorterms.py
parejkoj: Ok, I've left these as-is. I'll leave it to someone else to expand them if desired.

PR #356, opened 2021-04-12
DM-29615: include jointcal in DRP pipeline

TallJimbo:


pipelines/DRP.yaml
```diff
@@ -5,6 +5,18 @@ imports:
...
     class: lsst.pipe.drivers.skyCorrection.SkyCorrectionTask
+  jointcal:
+    # The jointcal package depends on obs_subaru, not the other way around, so
+    # it's not great to have this here, but it's the least-bad temporary spot
+    # for this information (which is specific to both Gen3 and HSC) until we
+    # RFC-approve a better home for it.
+    class: lsst.jointcal.JointcalTask
+    config:
+      doPhotometry: false
+      applyColorTerms: false  # only relevant for photometry
```
parejkoj, line 16: This line shouldn't be necessary: jointcal knows to only apply colorterms.

Looking at `obs_subaru/config/jointcal.py`, I see `config.applyColorTerms = True`, so this is overriding that override. Instead of making doPhotometry a gen3-specific thing in this file, I think we should instead change that jointcal config override file to disable photometry and remove the colorterms line.

pipelines/DRP.yaml
```diff
@@ -5,6 +5,18 @@ imports:
...
+      python: >
+        from lsst.utils import getPackageDir;
+        config.load(os.path.join(getPackageDir("jointcal"), "tests", "config", "config-gen3.py"));
```
parejkoj, line 19: I'd put a comment here about the test config being temporary until DM-29008.

pipelines/DRP.yaml
TallJimbo: Changing the jointcal config instead is a good idea; I'll go do that.

But jointcal does emit a warning when `applyColorTerms` is `True` but `doPhotometry` is `False`; squashing that warning was why I changed the setting here (and why I'll do the same in the config, unless you'd prefer I remove that warning from `jointcal`).

TallJimbo:
> I guess we can't do anything about that until we actually have a place to put these, but it feels very similar to the problem we had with HSC configs being the "actual" defaults without people knowing about them (when, e.g. running DECam for AP).

RFCing the better solution to this is the other thing I want to do _today_ (besides this ticket), so it won't be long.

pipelines/DRP.yaml
TallJimbo: So, I had forgotten briefly that jointcal is downstream of obs_subaru, and consumes obs_subaru's configs for its own tests. So moving this to the config turns off photometry in jointcal's tests, and that of course breaks them.  I think I've going to revert that entire "disable photometry" change for the purposes of this ticket and the weekly, and revisit it after the RFC I'm working on that would suggest reorganizing a lot of this configuration anyway.

parejkoj: applyColorTerms defaults to False (after all, most instruments+refcat combos don't have colorterms), so all you need to do is remove the line setting it to True in obs_subaru.

parejkoj: To fix the jointcal tests, just add `config.doPhotometry=True` to `tests/config/hsc-config.py`

TallJimbo: Ah, that does sound pretty easy.  I'm going to let my current Jenkins run complete, then try again with disabling photometry in the way you suggested after squashing the rest of my post-review changes, so if that doesn't go smoothly I can remove those commits and still make the weekly.

TallJimbo: I'm bailing out on this after all; some jointcal tests are still failing and I've got other tickets to merge tonight, so I can't afford another long Jenkins run for this one.  I'll open a new ticket to turn off photometry later.

PR #363, opened 2021-05-26
DM-30342: Add missing detect flags to object table

fred3m:


jeffcarlin (approved):
You will want to add a renaming rule at the bottom like ['merge_', 'Merge_'] to properly deal with `merge_peak_sky`. Otherwise, all looks good.

yalsayyad (commented):
Because the "merge_*" flags are from the reference catalog, they are not exploded by filter and therefore should not start with a capital letter. I would actually remove the rename rule for merge  unless you want a those columns to be the only columns in the table that start with a capital letter. Ew.

yalsayyad:
Actually if the code were smart, it would be a no op. Let's see...

PR #367, opened 2021-06-23
DM-30858: Add GAaP to obs_subaru

arunkannawadi:


config/charImage.py
```diff
@@ -38,6 +38,7 @@
 config.measurement.load(os.path.join(ObsConfigDir, "apertures.py"))
 config.measurement.load(os.path.join(ObsConfigDir, "kron.py"))
 config.measurement.load(os.path.join(ObsConfigDir, "convolvedFluxes.py"))
+config.measurement.load(os.path.join(ObsConfigDir, "gaap.py"))
```
yalsayyad, line 41: I don't think we  want to run gaap in charImage. charImage is not what makes the `src` catalogs, and I don't think we want to run it in `calibrate` (which does make the `src` catalogs) either. We don't run any galaxy measurements in single frame processing.

config/forcedPhotCcd.py
```diff
@@ -2,6 +2,7 @@
 
 config.measurement.load(os.path.join(os.path.dirname(__file__), "apertures.py"))
 config.measurement.load(os.path.join(os.path.dirname(__file__), "kron.py"))
+config.measurement.load(os.path.join(os.path.dirname(__file__), "gaap.py"))
```
yalsayyad, line 5: Don't turn on here:
1) no slow measurement plugin on any per-visit processing. 
2) Even if it were fast, I'm not sure it makes sense. 
Let's talk about it first.

config/gaap.py
```diff
@@ -0,0 +1,7 @@
+# Enablue GAaP (Gaussian Aperture and PSF) colors
+# 'config' is a SourceMeasurementConfig
```
yalsayyad, line 2: This comment isn't universally true

config/gaap.py
```diff
@@ -0,0 +1,7 @@
+# Enablue GAaP (Gaussian Aperture and PSF) colors
```
yalsayyad, line 1: Enablue to Enable

config/measureCoaddSources.py
```diff
@@ -4,6 +4,7 @@
 config.measurement.load(os.path.join(os.path.dirname(__file__), "kron.py"))
 config.measurement.load(os.path.join(os.path.dirname(__file__), "convolvedFluxes.py"))
 config.measurement.load(os.path.join(os.path.dirname(__file__), "hsm.py"))
+config.measurement.load(os.path.join(os.path.dirname(__file__), "gaap.py"))
```
yalsayyad, line 7: Don't turn on here.

policy/Object.yaml
```diff
@@ -150,6 +150,50 @@ funcs:
         args: slot_ModelFlux
         filt1: z
         filt2: y
+    #  PSF GAaP flux
+    GaapPsFlux:
+        functor: NanoJansky
+        dataset: meas
```
yalsayyad, line 156: `dataset: forced_src` if we're running only on forcedPhotCoadd.

arunkannawadi:
I got a comment on my email about not wanting to run gaap in `config/charImage.py` that I don't find it here - assuming you took it back? That was there to register for aperture corrections, not running the measurement itself.

PR #377, opened 2021-08-05
DM-31324: add step4 and step5

yalsayyad:


pipelines/DRP.yaml
```diff
@@ -199,3 +202,38 @@ subsets:
...
+    description: >
+        Tasks that can be run together, but only after the 'step1', 'step2',
+        'step3', and 'step4' subsets
+
+        These DIA aggregation steps include both visit-level aggregation and
+        tract-level aggregation drpAssociation and drpDiaCalculation.
+
+        If run with tract or patch in the data query, drpAssociation and
+        drpDiaCalculation will be complete, but the outputs of
+        consolidateDiaSourceTable not complete visits.
```
erykoff, line 233: Like obs_lsst, I don't understand this sentence, it's missing a verb. Also, should this be run with a tract or patch or without a tract or patch? That's not clear.

erykoff (approved):
See comment on step5 description.

pipelines/DRP.yaml
```diff
@@ -199,3 +202,38 @@ subsets:
...
+
+        This subset is separate to signal to operators to hold to assess
```
laurenam, line 235: I got a bit confused with the meaning of "hold to assess".  Perhaps "pause to assess" instead?

PR #379, opened 2021-08-19
DM-31462: Add randomSeed to fgcm build stars for repeatability.

erykoff:


pipelines/DRP.yaml
```diff
@@ -36,6 +36,7 @@ tasks:
       densityCutMaxPerPixel: 1500
       primaryBands: ['i', 'r', 'g', 'z', 'y', 'N921']
       doReferenceMatches: True
+      randomSeed: 23456
```
laurenam, line 39: Just making double sure you specifically do not want to add this to https://github.com/lsst/obs_subaru/blob/master/config/fgcmBuildStarsTable.py?

laurenam (approved):

pipelines/DRP.yaml
erykoff: Yeah, I don't actually want a random seed to be a global default...

PR #381, opened 2021-08-24
DM-31489: Update StrayLightData to use FitsGenericFormatter with a deferred data set

czwa:
This change makes the `SubaruStrayLightDataFormatter` a subclass of `FitsGenericFormatter`, and implements a `readFits` method to do so.  The reads differ between the butler generations:

* Gen2: `IsrTask.readIsrData` calls `SubaruStrayLightTask.readIsrData`, which checks if the exposure needs correcting, and if so, calls the `SubaruStrayLightData.readFits` method.  `SubaruStrayLightData.run` is called in the standard point in `IsrTask.run`.
* Gen3: A `DeferredDatasetHandle` is passed to `IsrTask.run` which calls `SubaruStrayLightData.run` at the standard point. The `run` method calls the same check method as in Gen2 (indicating this is run twice in Gen2), and if a correction is needed, converts the handle into the real object.

There are some slight naming clarifications (`hdulist` -> `calib.ampData`) and the docstrings in edited files have been converted to DM standards.

### DIFF
--- a/python/lsst/obs/subaru/strayLight/yStrayLight.py
+++ b/python/lsst/obs/subaru/strayLight/yStrayLight.py
@@ -33,7 +33,6 @@ from .rotatorAngle import inrStartEnd
 BAD_THRESHOLD = 500  # Threshold for identifying bad pixels in the reconstructed dark image
 
 
-# TODO DM-16805: This doesn't match the rest of the obs_subaru/ISR code.
 class SubaruStrayLightTask(StrayLightTask):
     """Remove stray light in the y-band
 
@@ -49,6 +48,7 @@ class SubaruStrayLightTask(StrayLightTask):
     This Task retrieves the appropriate dark, uncompresses it and
     uses it to remove the stray light from an exposure.
     """
+
     ConfigClass = StrayLightConfig
 
     def readIsrData(self, dataRef, rawExposure):
@@ -141,7 +141,7 @@ class SubaruStrayLightTask(StrayLightTask):
 
 
 class SubaruStrayLightData(StrayLightData):
-    """Lazy-load object that reads and integrates the wavelet-compressed
+    """Object that reads and integrates the wavelet-compressed
     HSC y-band stray-light model.
 
     Parameters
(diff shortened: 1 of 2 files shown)


python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -150,8 +155,20 @@ class SubaruStrayLightData(StrayLightData):
...
-    def __init__(self, filename):
-        self._filename = filename
+    @classmethod
+    def readFits(cls, filename, **kwargs):
+        calib = cls()
+        hdulist = fits.open(filename)
+
+        calib.ampData = []
+        for extension in (0, 1, 2, 3):
+            calib.ampData.append(hdulist[extension].data)
```
timj, line 165:
```python
with fits.open(filename) as hdulist:
    calib.ampData = [hdulist[extension].data for extension in range(4)]
    calib.setMetadata(hdulist[0].header)
```
?

python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -150,8 +155,16 @@ class SubaruStrayLightData(StrayLightData):
...
     """
 
-    def __init__(self, filename):
-        self._filename = filename
+    @classmethod
+    def readFits(cls, filename, **kwargs):
+        calib = cls()
+
+        with fits.open(filename) as hdulist:
+            calib.ampData = [hdulist[extension].data for extension in range(4)]
```
timj, line 163: Looks like previously the code use `len(hdulist)` so maybe that should be done here as well?
```python
calib.ampData = [hdu.data for hdu in hdulist]
```
? (I think `hdulist` can be an iterator directly).

python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -175,20 +188,19 @@ def evaluate(self, angle_start: Angle, angle_end: Optional[Angle] = None):
         ccd_img : `numpy.ndarray`
             Background data for this exposure.
         """
-        hdulist = fits.open(self._filename)
-        header = hdulist[0].header
+        header = self.getMetadata().toDict()
```
timj, line 191: If `getMetadata()` returns a PropertyList is the `toDict()` really necessary? Doesn't `header["F_NAXIS2"]` already work fine? or are there some multi-value headers in play?

PR #382, opened 2021-08-30
DM-31580: Move consolidateDiaSourceTable to step6

hsinfang:


yalsayyad (approved):
See comments on the obs_lsst PR

PR #388, opened 2021-09-22
DM-31861: Copy overrides from jointcal to DRP.yaml

SimonKrughoff:


pipelines/DRP.yaml
```diff
@@ -21,9 +21,31 @@ tasks:
       doPhotometry: false
       applyColorTerms: false
       python: |
-        # This override is temporary until DM-29008.
-        from lsst.utils import getPackageDir
-        config.load(os.path.join(getPackageDir("jointcal"), "tests", "config", "config-gen3.py"))
```
erykoff, line 26: I think it might be worth adding a comment on where these variables came from.

erykoff (approved):

pipelines/DRP.yaml
erykoff: We should have a TO-DO on DM-29008 that once that is done then these overrides are no longer necessary.

SimonKrughoff: You mean just saying "These defaults were copied from the config-gen3.yaml file provided in the tests/config directory of the jointcal package"?

erykoff: Yup, and the TODO.

PR #393, opened 2021-10-13
DM-32029: Add faro steps to DRP.yaml

jeffcarlin:
Successful Jenkins run is at https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/35175/pipeline/

The blocker (DM-32059) has been completed and merged - that ticket removed the unused `faro` tasks from the main pipelines so they wouldn't cause problems in `obs_subaru`.

### DIFF
--- a/pipelines/DRP.yaml
+++ b/pipelines/DRP.yaml
@@ -231,7 +231,9 @@ subsets:
       - fgcmBuildStarsTable
       - fgcmFitCycle
       - fgcmOutputProducts
-      - faro_step2
+      - nsrcMeasVisit
+      - TE3
+      - TE4
     description: >
       Per-visit tasks that can be run together, but only after the 'step1'.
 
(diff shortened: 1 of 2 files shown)


pipelines/DRP.yaml
```diff
@@ -160,6 +161,45 @@ subsets:
       The multiVisit subset defined in pipe_tasks' DRP.yaml is not safe to
       use on HSC for various reasons; use 'step1', 'step2', and 'step3' subsets
       instead.  It may be re-enabled in the future.
+  faro:
```
arunkannawadi, line 164: Would it make sense to categorize faro tasks as faro_step1, faro_step2 etc. based on when they can be run and faro_step1 can be included in step1? If that is not possible to merge before the weekly deadline, please go ahead and merge but you could consider reorganization later. The reason I ask is that it'd make it easier to add new metrics/tasks later on as we add them.

arunkannawadi (commented):
On the commits themselves - why are there two sets of commits with identical messages? There probably shouldn't be the merge commit either. And commit messages should not end in a period

jeffcarlin:
> On the commits themselves - why are there two sets of commits with identical messages? There probably shouldn't be the merge commit either. And commit messages should not end in a period

I think something got funky when I rebased. We can squash those commit messages before merging.

arunkannawadi (approved):

arunkannawadi:
Ah, this was the part I was unsure about and wanted Jeff to check. Good to
know that this isn't possible right now.

arunkannawadi:
What you have seems to be a good compromise between what is possible and what would have been ideal (but impossible). And you seem to have picked up some other version of my commits from yesterday which are already merged onto `master`. Have you rebased it correctly? Please checkout `master` branch, pull all the changes, and rebase it to the latest `master`. Drop the merge commits and any other commits that have to do with GAaP

PR #403, opened 2021-11-30
DM_31964: Fixup dimensions on ProcessCcdWithFakesTask/MatchFakesTask

morriscb:


pipelines/DRPFakes.yaml
```diff
@@ -35,7 +35,7 @@ imports:
     - TE2
     - wPerp
 tasks:
-  processCcdWithFakes:
+  detectorVisitFakes:
```
leeskelvin, line 38: This new name doesn't seem to capture what this task does. Would singleFrameWithFakes work better here?

PR #407, opened 2022-01-10
DM-27013: Change CalibrateTask refcat defaults to Gaia DR2 for astrometry and PS1 for photometry

parejkoj:


parejkoj:
Closing this: HSC will explore the effect of the Gaia astrometry change independently, so that this doesn't hold up the rest of the ticket.

PR #412, opened 2022-03-03
DM-28334: Replace afw.image.Filter name usage with explicit map.

TallJimbo:
We have a few data files committed to git here that use various old names for filters.  It's best to just have an explicit constant mapping for those.

### DIFF
--- a/python/lsst/obs/hsc/makeTransmissionCurves.py
+++ b/python/lsst/obs/hsc/makeTransmissionCurves.py
@@ -27,7 +27,6 @@ import numpy as np
 
 from lsst.afw.image import TransmissionCurve
 from lsst.utils import getPackageDir
-from . import hscFilters
 
 __all__ = ("getOpticsTransmission", "getSensorTransmission", "getAtmosphereTransmission",
            "getFilterTransmission",)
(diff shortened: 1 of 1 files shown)


python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -36,20 +35,23 @@
...
 
+OLD_FILTER_NAME_MAP = {b: f"HSC-{b.upper()}" for b in "grizy"}
+OLD_FILTER_NAME_MAP.update(r2="HSC-R2", i2="HSC-I2")
 
-def getLongFilterName(short):
-    """Return a long HSC filter name (e.g. 'HSC-R') that's usable as a data ID
-    value from the short one (e.g. 'r') declared canonical in afw.image.Filter.
+
+def getPhysicalFilterName(short):
+    """Return am HSC physical_filter name (e.g. 'HSC-R') that's usable as a
```
parejkoj, line 43: "am"->"an"

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -36,20 +35,23 @@
...
     if short.startswith("HSC"):
         return short
     if short.startswith("NB") or short.startswith("IB"):
         num = int(short[2:].lstrip("0"))
         return "%s%04d" % (short[:2], num)
-    for filter in hscFilters.HSC_FILTER_DEFINITIONS:
-        if short == filter.afw_name or short == filter.band:
-            return filter.physical_filter
-    return short
+    if short.startswith("N"):
```
parejkoj, line 51: Might be worth a comment above this that it's just here for the narrowband filters?

python/lsst/obs/hsc/makeTransmissionCurves.py
```diff
@@ -36,20 +35,23 @@
 
 HSC_BEGIN = "2012-12-18"  # initial date for curves valid for entire lifetime of HSC
 
+OLD_FILTER_NAME_MAP = {b: f"HSC-{b.upper()}" for b in "grizy"}
+OLD_FILTER_NAME_MAP.update(r2="HSC-R2", i2="HSC-I2")
```
parejkoj, line 39: Should we have a comment above this saying something like "this is to map physical filter names to the deprecated afw names", for future readers?

parejkoj (commented):
I think my only real concern is whether this code is actually run anywhere, since I thought the transmission curves that we use are pre-built and just exist in obs_subaru_data?

timj:
Also:

> we have a few data files committed to git here

Can't we fix the files?

TallJimbo:
> I think my only real concern is whether this code is actually run anywhere, since I thought the transmission curves that we use are pre-built and just exist in obs_subaru_data?

Yes, these are still run: `HyperSuprimeCam.writeAdditionalCuratedCalibrations` calls the functions in `makeTransmissionCurves.py` to assemble `TransmissionCurve` instances from the data files in `obs_subaru/hsc/transmission`, and then writes those to the data repository.

> Also:
>>
>>    we have a few data files committed to git here
>>
> Can't we fix the files?

I just don't think it's worth it.

PR #414, opened 2022-03-18
DM-33933: Apply Extinction Correction to colorColorPlot and colorColorFitPlot

yalsayyad:
- Rename existing coeffs file with old Filter as the dict key to
  extinctionCoeffsGen2.py. It is still being used in pipe_analysis.

### DIFF
--- a/config/extinctionCoeffs.py
+++ b/config/extinctionCoeffs.py
@@ -2,13 +2,17 @@
 # Numbers provided by Masayuki Tanaka (NAOJ).
 #
 # Band, A_filter/E(B-V)
+
+# These default extinction coefficients are not optimal for the FGCM standard bandpasses.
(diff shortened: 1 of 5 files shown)


config/extinctionCoeffs.py
```diff
@@ -2,13 +2,17 @@
 # Numbers provided by Masayuki Tanaka (NAOJ).
 #
 # Band, A_filter/E(B-V)
+
+# These default extinction coefficients are WRONG, but better than nothing
```
erykoff, line 6: I think that all-caps WRONG is a bit strong here.  "These default extinction coefficients are not optimal for the FGCM standard bandpasses.  TO DO..."

erykoff (approved):

PR #416, opened 2022-04-20
DM-27013: Change CalibrateTask refcat defaults to Gaia DR2 for astrometry and PS1 for photometry

parejkoj:


ktlim (approved):
Looks like the same thing you have in, e.g., obs_lsst.

ktlim:
Oh, but adding in the comment: "# Use the filterMap instead of the "any" filter." might be good, too.

PR #418, opened 2022-05-19
DM-34856: Remove gen2 code from obs_subaru.

erykoff:
This also updates example plotting code to gen3 where it is straightforward, and removes code that is heavily gen2 and assumed to be unused.


tests/test_distortion.py
```diff
@@ -49,7 +48,7 @@ def testDistortion(self):
         data
 
         If SAVE_DATA is true then save newly created data and then fail the
-        test in order to prevent anyone from committing the test with SAVE_DATA
+        test in order to prevent anyoe from committing the test with SAVE_DATA
```
timj, line 51: Typo

arunkannawadi (approved):
Is not everything in bin.src/ related to Gen2? I haven't checked if the changes are exhaustive, but the code deletion looks good.

erykoff:
In bin.src are some routines for doing plotting, I guess they're closer to example code than anything else.  And the plotting examples do run after my updates.  (Having a camera plotting bit is useful!)

erykoff:
You're right about the remaining things in bin.src, I've removed the old ones.

PR #425, opened 2022-07-13
DM-33429: Add ability to do both serial and parallel overscan correction

czwa:
Switch config file to use current overscan config options instead of deprecated isrTask config options.

### DIFF
--- a/config/isr.py
+++ b/config/isr.py
@@ -26,13 +26,6 @@ config.doOverscan = True
 config.overscan.fitType = "AKIMA_SPLINE"
 config.overscan.order = 30
 config.overscan.numSigmaClip = 3.0
-config.overscanNumLeadingColumnsToSkip = 0
-config.overscanNumTrailingColumnsToSkip = 0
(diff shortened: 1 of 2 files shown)


config/isr.py
```diff
@@ -26,13 +26,9 @@
...
-config.overscanNumLeadingColumnsToSkip = 0
-config.overscanNumTrailingColumnsToSkip = 0
-config.overscanMaxDev = 1000.0
-config.overscanBiasJump = False
-config.overscanBiasJumpKeyword = "NO_SUCH_KEY"
-config.overscanBiasJumpDevices = "NO_SUCH_KEY"
-config.overscanBiasJumpLocation = -1
+config.overscan.leadingColumnsToSkip = 0
+config.overscan.trailingColumnsToSkip = 0
+config.overscan.maxDeviation = 1000.0
```
PaulPrice, line 31: Aren't these the defaults? Then can't we just remove them?

PaulPrice (approved):

PR #430, opened 2022-08-05
DM-33978: Remove charImage.py and move content to characterizeImage.py

timj:


parejkoj (approved):
Thanks for cleaning this up.

PR #431, opened 2022-08-08
DM-35670: Remove deprected config overrides.

erykoff:


timj (approved):
Looks okay. Feel free to also remove `config/*Driver.py` and `config/processCcd*.py` files that are now obsolete.

PR #445, opened 2023-05-04
DM-38561: remove Gen2 ingest and codes importing it 

aferte:


timj (approved):
Thanks. Sorry I forgot to remove them previously.

PR #456, opened 2023-07-06
DM-39227: drop overrides for deprected config options

TallJimbo:


config/forcedPhotCcd.py
```diff
@@ -6,6 +6,3 @@
 config.measurement.plugins["base_CircularApertureFlux"].radii = [12.0]
 
 config.measurement.slots.gaussianFlux = None
-config.doApplyExternalPhotoCalib = True
-config.doApplyExternalSkyWcs = True
-config.doApplySkyCorr = True
```
leeskelvin, line 11: Are `skyCorr` datasets also being bundled into `finalVisitSummary` tables? From looking through the `forcedPhotCcd` script in the [meas_base PR](https://github.com/lsst/meas_base/pull/248/files), I don't think so?

config/coaddBase.py
```diff
@@ -1,4 +1,2 @@
 # Configs shared between makeWarp and assemble
 config.matchingKernelSize = 29
-config.doApplyExternalPhotoCalib = True
-config.doApplyExternalSkyWcs = True
```
leeskelvin, line 4: By removing these `doApply...` configs (here and elsewhere), are we also removing the flexibility to turn these components off? I see a new `useVisitSummaryPsf` config option - do we also need similar `useVisitSummaryPhotoCalib` and `useVisitSummaryWcs` configs to replicate this functionality?

TallJimbo: No, we are intentionally dropping this functionality _here_, because the expectation is that controlling what goes into a visit summary dataset (which still has these knobs) is sufficient for everything downstream of that visit summary and upstream of any other visit summary.

config/forcedPhotCcd.py
TallJimbo: Ah, good catch.  Backgrounds are not in visit summaries, and I think removing this last line was a mistake.

config/coaddBase.py
leeskelvin: Got it, thanks for the explanation, good to know!

leeskelvin (approved):

PR #461, opened 2023-09-26
DM-40755: Update default maximum number of iterations for fgcm fit.

erykoff:


laurenam (approved):

config/fgcmFitCycle.py
```diff
@@ -5,7 +5,7 @@
 config.fitBands = ['N387', 'g', 'r', 'i', 'N816', 'z', 'N921', 'y', 'N1010']
 from lsst.obs.hsc.hscFilters import HSC_FILTER_DEFINITIONS
 config.physicalFilterMap = HSC_FILTER_DEFINITIONS.physical_to_band
-config.maxIterBeforeFinalCycle = 75
+config.maxIterBeforeFinalCycle = 150
```
laurenam, line 8: If you're not using the updated LUT for processing (on the ticket you indicated that's TBD), don't you need to add `config.fitCcdChromaticityDict = {'g': True}` here as well?  (Or I'm still missing something!)

PR #466, opened 2024-01-27
DM-42080: Add effTime to computeExposureSummaryStats

kadrlica:


config/updateVisitSummary.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+import os.path
+
+ObsConfigDir = os.path.dirname(__file__)
+
+print("-----DEBUG: config/updateVisitSummary.py-----")
```
erykoff, line 6: Please remove debug statement.  Thanks!

erykoff (changes requested):

config/updateVisitSummary.py
kadrlica: Oops! Should be fixed.

erykoff (approved):

PR #474, opened 2024-06-04
DM-44623: Remove matchingKernelSize config override

arunkannawadi:


arunkannawadi:
This is not for this ticket but DM-44688 or the corresponding removal ticket.

PR #480, opened 2024-07-02
DM-45068: configure ruff/flake8 to lint subaru configs and camera

parejkoj:


config/gaap.py
```diff
@@ -2,9 +2,10 @@
 # 'config' is typically a SourceMeasurementConfig
 try:
     import lsst.meas.extensions.gaap  # noqa
+
     config.plugins.names.add("ext_gaap_GaapFlux")
     config.plugins["ext_gaap_GaapFlux"].sigmas = [0.5, 0.7, 1.0, 1.5, 2.5, 3.0]
-    # Enable PSF photometry after PSF-Gaussianization in the `ext_gaap_GaapFlux` plugin
+    # Enable PSF photometry after PSF-Gaussianization for  `ext_gaap_GaapFlux`
```
erykoff, line 8: This has an extra space

setup.cfg
```diff
@@ -1,7 +1,14 @@
+# NOTE: this package is now linted by ruff in GitHub Actions and scons;
+# this section is retained here for compatibility with users who still use
+# flake8 in their local workflows, but may not produce exactly the same
+# results as the ruff config.
```
erykoff, line 4: This phrasing is a bit awkward.  It is written as though the *users* may not produce the same results.  Something like: "This section is retained here for best-effort compatibility with flake8 for users who do not use ruff in their local workflows.  Note, however, that this may not produce exactly the same results as the ruff config which is required to pass before a PR can be accepted."

erykoff (approved):
A couple of minor comments

setup.cfg
parejkoj: Thanks, yeah, that's better. I'll make the same change to cfht, decam, and lsst (the latter of which was done on a different ticket).

PR #481, opened 2024-08-01
DM-45507: Split amp offset config into measurement and application components

enourbakhsh:


config/isr.py
```diff
@@ -105,7 +105,9 @@
 
 config.doNanInterpAfterFlat = False
 
-config.doAmpOffset = True
+# Calculate and apply amp offset corrections.
+config.doMeasureAmpOffset = True
```
czwa, line 109: I think I'm about to comment on the other PR that these options should be `config.ampOffset.doMeasureAmpOffset`

enourbakhsh: I thin you meant `config.ampOffset.doApplyAmpOffset`?

enourbakhsh: Done!

czwa (approved):

PR #482, opened 2024-08-02
DM-39168: Add objectEpochTable bands

cmsaunders:


config/objectEpochTable.py
```diff
@@ -0,0 +1,8 @@
+import os.path
+from lsst.pipe.tasks.postprocess import TransformObjectCatalogConfig
```
parejkoj, line 2: This method of importing another config to load a different config file isn't something I've seen before.

I think I understand what it's doing, but this is all suggesting to me that we should really just set ugrizy as the default `outputBands` in the base `TransformObjectCatalogConfig`, since that's what we want for LSST.

cmsaunders: Yeah, this is used a lot in `obs_subaru` to automatically grab the band names so they don't have to be hardcoded. And yes, the `outputBands` should probably be ugrizy by default in `TransformObjectCatalogConfig`.

parejkoj: Please make a ticket to change `outputBands` default to be ugrizy and remove all of our overrides of it (unless they're something different).

cmsaunders: Looking at `TransformObjectCatalogConfig` again, I don't think we want to set this. The default mode is for `outputBands` to be the set by the `inputBands`, so it is often appropriate for it to be `None`. If we add specific bands there, it's going to mess up expected behavior.

PR #483, opened 2024-08-25
DM-45878: Split out obs_package fiducial configs into their own files for use in multiple tasks

laurenam:


kadrlica (approved):
Looks good.

PR #484, opened 2024-08-28
DM-42576: mpSkyEphemerisQuery config overrides

Gerenjie:
Just adding config overrides for mpSkyEphemerisQuery -- observatory code T09 and a 0.2 degree buffer radius for our search.

### DIFF
--- a/config/mpSkyEphemerisQuery.py
+++ b/config/mpSkyEphemerisQuery.py
@@ -1,2 +1,1 @@
 config.observerCode = "T09"  # IAU code for Subaru
-config.queryBufferRadiusDegrees = 0.2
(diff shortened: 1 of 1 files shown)


config/mpSkyEphemerisQuery.py
```diff
@@ -0,0 +1,2 @@
+config.observerCode = "T09"  # IAU code for Subaru
+config.queryBufferRadiusDegrees = 0.2
```
kfindeisen, line 2: Is there any need to override `queryBufferRadiusDegrees`, if it's just set by proper motions?

kfindeisen (approved):
Looks good, but I don't think the buffer needs to be customized.

PR #501, opened 2025-03-25
DM-49673: Make spatialOrder config in piff per-band and add a minNPsfStar criterion for detector rejection

laurenam:


config/finalize_characterization.py
```diff
@@ -0,0 +1 @@
+config.psf_determiner['piff'].spatialOrderPerBand = {"fallbadk": 2}
```
PFLeget, line 1: there is a typo:

```suggestion
config.psf_determiner['piff'].spatialOrderPerBand = {"fallback": 2}
```

PFLeget (approved):

config/finalize_characterization.py
laurenam: Good catch!  Fixed and checked:

```
$ pipetask build -p $DRP_PIPE_DIR/pipelines/HSC/DRP-RC2.yaml#finalizeCharacterizationDetector --show config="*piff*patialOr*"
Matching "*piff*patialOr*" without regard to case (append :NOIGNORECASE to prevent this)
### Configuration for task `finalizeCharacterizationDetector'
# Per-band spatial order for PSF kernel creation. Ignored if piffPsfConfigYaml is set.
config.psf_determiner['piff'].spatialOrderPerBand={'fallback': 2}
```

PR #502, opened 2025-03-27
DM-49727: Add bands override for ComputeObjectEpochsTask

taranu:


TallJimbo (approved):
Approved, though I'm hoping it works to drop the `bands` config so this can go away.

PR #503, opened 2025-03-28
DM-49576: Compare DateTimes for straylight

timj:
More efficient than converting to two python datetime objects.

### DIFF
--- a/python/lsst/obs/subaru/strayLight/yStrayLight.py
+++ b/python/lsst/obs/subaru/strayLight/yStrayLight.py
@@ -16,13 +16,13 @@
 
 __all__ = ["SubaruStrayLightTask"]
 
-import datetime
 from typing import Optional
 
 import numpy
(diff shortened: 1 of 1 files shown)


python/lsst/obs/subaru/strayLight/yStrayLight.py
```diff
@@ -33,6 +33,9 @@
 
 BAD_THRESHOLD = 500  # Threshold for identifying bad pixels in the reconstructed dark image
 
+# Fixed on 2018-01-01
+_STRAY_LIGHT_FIXED_MJD = 58119.0
```
ktlim, line 37: This is not as self-documenting or user-friendly.  It encourages a floating point comparison, although an exact one.  Might it be better to convert 2018-01-01T00:00:00 to a `lsst.daf.base.DateTime` and compare that?

ktlim (approved):

PR #508, opened 2025-07-23
DM-51429: Add config override to bgModel2.minFrac

leeskelvin:
On DM-51429, the task default for bgModel2.minFrac was changed to 0.3. This change was made to facilitate skyCorr processing for LSSTCam data, whereby crowded fields make the existing default of 0.5 too high. To preserve HSC behaviour, an HSC-specific config override is set here to maintain a value of bgModel2.minFrac=0.5.


PFLeget (approved):
lgtm

PR #509, opened 2025-08-29
DM-52355: Add support for NB0395 and NB0515

leeskelvin:


config/colorterms.py
```diff
@@ -18,6 +18,7 @@
         'HSC-Y': Colorterm(primary="z", secondary="i", c0=0.003386, c1=0.428877, c2=0.076738),
         'IB0945': Colorterm(primary="z", secondary="i", c0=0.008117, c1=0.234991, c2=-0.042255),
         'NB0387': Colorterm(primary="u", secondary="g", c0=-0.709229, c1=0.310719, c2=-0.044107),
+        'NB0395': Colorterm(primary="u", secondary="g", c0=-0.484711, c1=-0.095468, c2=0.002404),
```
taranu, line 21: Were these color terms derived the same way as the others? I have a vague recollection of someone being unsure of the provenance of some of them.

config/isolatedStarAssociation.py
```diff
@@ -1,2 +1,2 @@
 # irgzyN921 is preferred order of Eli and tested in RC2
-config.band_order = ["i", "r", "g", "z", "y", "N921", "N816", "N1010", "N387", "N515"]
+config.band_order = ["i", "r", "g", "z", "y", "N921", "N816", "N1010", "N387", "N515", "N395"]
```
taranu, line 2: I doubt it matters much but in practice is there some logic to the order of the narrow bands here?

taranu (approved):
Seems fine; I just have questions for clarification.

config/colorterms.py
leeskelvin: These color terms are taken verbatim from the hscPipe 8.x patch script circulated amongst the group in Japan (attached to the Jira ticket). I've spot checked a number of the bands that we've already defined and those that are similarly defined in the patch script, and those numbers are identical, so I'm fairly confident that these numbers should be appropriate here too.

config/isolatedStarAssociation.py
leeskelvin: I'm not sure. I know in other files the order is important (e.g., [hscFilters.py](https://github.com/lsst/obs_subaru/blob/5527664d53b3ba509dc570056a3bfb45bf47ec75/python/lsst/obs/hsc/hscFilters.py)), so I tried to match prior art as closely as possible. Where bands appeared in wavelength order, I slotted the new band(s) in; where they appeared in a seemingly arbitrary order, I appended to the end, as here.

